PR #11, opened 2026-05-19
Feature: add API for fetching node info

### LINKED ISSUE #3: Efficient Cross-Node Metrics Collection
Conduct an in-depth investigation into retrieving system statistics (memory, CPU, process counts, port limits) from a monitored Node (A) to the Beholder Node (B). We must identify the method with the absolute lowest overhead on the production system.

- Analyze low-level functions (:erlang.memory/0, :erlang.statistics/1, etc.) and identify which are the most "production-safe" under high load.
- Deconstruct the implementations found in [Phoenix LiveDashboard](https://github.com/phoenixframework/phoenix_live_dashboard), [observer](https://www.erlang.org/doc/apps/observer/observer.html) and [observer_web](https://github.com/thiagoesteves/observer_web) to find their performance bottlenecks.
- Evaluate the trade-offs between polling (request/response) and push-based telemetry.
- Determine how to handle nodes with a massive number of processes (100k+) without blocking the scheduler or saturating the network.

Expected Outcome:
A technical proposal (Markdown) outlining the data collection strategy. It must be "production-ready" and avoid any shortcuts that could lead to node instability.

Decide if this is possible to achieve this fetching without adding our dependency to debugged node. There is a chance that it may be better to provide some functions to the debugged node via our dependency (e.g. beholder_backend) and then just call them via `:erpc` from beholder node

kraleppa:
I can't mark all of you as reviewers so I mark you here: @srzeszut @GuzekAlan @hhubert6


lib/voyager/services/node_info/language.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  @type t :: %__MODULE__{
+          name: String.t(),
+          version: String.t()
+        }
+
+  defstruct [:name, :version]
+
+  @spec candidate_apps() :: [atom()]
+  def candidate_apps, do: Enum.map(@applications, fn {app, _name} -> app end)
```
srzeszut:
```suggestion
  @candidate_apps Enum.map(@applications, fn {app, _name} -> app end)
  
  @spec candidate_apps() :: [atom()]
  def candidate_apps, do: @candidate_apps
```
maybe like this?

srzeszut (approved):
Nice!

lib/voyager/services/node_info/memory.ex
```diff
@@ -0,0 +1,59 @@
...
+    %__MODULE__{
+      total: total,
+      processes_allocated: process_allocated,
+      processes_used: process_used,
+      atom_allocated: atom_allocated,
+      atom_used: atom_used,
+      binary: binary,
+      code: code,
+      ets: ets,
+      other: max(total - (process_allocated + atom_allocated + binary + code + ets), 0)
```
GuzekAlan, line 56: Is total being lower then sum of rest something common ? If not then what do you think about logging that it happened ?

GuzekAlan (approved):
LGTM !

lib/voyager/services/node_info/memory.ex
kraleppa: it's theoretically possible for the named categories to temporarily exceed :total due because the values aren't read atomically.

It's very rare, but still this max is a small defensive guard. I wouldn't log it since this is not an anomaly

PR #14, opened 2026-05-22
Feat: add exporting telemetry from voyager

### LINKED ISSUE #8: App usage telemetry
Collecting info about uses of Voyager from users to have better insight

GuzekAlan:
I'll tag you here: @kraleppa @hhubert6 @srzeszut


dev/prod_with_secrets.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# Simple script to run the server in production mode. `.env` leaked to test sandbox.
+
+env=prod
+database_path=priv/db/voyager_prod.db
+telemetry_push_url=http://127.0.0.1:4310/telemetry
+secret_key_base=EoaZBjLjm/W+OmiaGW7kY6MP204JuaVG5Yl7G1Q98AZdgno+4XM36EnW3rpI8o3u
```
kraleppa, line 7: Are we sure that we want to keep secret_key_base for (even for dev) in a public repo? Can't we generate it via ` mix phx.gen.secret`

GuzekAlan: This is a script for easy setup prod with proper envs for development (to test Telemetry on production)

lib/voyager/telemetry.ex
```diff
@@ -0,0 +1,80 @@
...
+  String name is the dotted event name, e.g. `"voyager.node.connect"`.
+
+  Node connection events require `metadata: %{id: unique_identifier}` where
+  `id` is a non-empty string that identifies the connection session.
+  """
+  @spec dispatch(String.t(), Keyword.t()) ::
+          :ok | {:error, :unknown_event | :missing_identifier}
+  def dispatch(event_name, opts \\ [])
+
+  def dispatch("voyager.node.connect", []) do
```
kraleppa: Are we sure that we want to match here with this empty list? 
```suggestion
  def dispatch("voyager.node.connect", opts) do
```

lib/voyager/telemetry.ex
```diff
@@ -0,0 +1,80 @@
...
+    Supervisor.init(children, strategy: :one_for_one)
+  end
+
+  @doc """
+  Dispatch an event by string name.
+
+  String name is the dotted event name, e.g. `"voyager.node.connect"`.
+
+  Node connection events require `metadata: %{id: unique_identifier}` where
+  `id` is a non-empty string that identifies the connection session.
```
kraleppa: Docs outdated

lib/voyager/telemetry/handler.ex
```diff
@@ -0,0 +1,42 @@
...
+  @impl GenServer
+  def init(opts) do
+    handler_module =
+      case Keyword.fetch!(opts, :telemetry_handler) do
+        :export -> Voyager.Telemetry.Handler.Export
+        :logger -> Voyager.Telemetry.Handler.Logger
+        _ -> Voyager.Telemetry.Handler.Noop
+      end
+
+    :telemetry.attach_many(@handler_id, Events.events(), &handler_module.handle_event/4, %{})
```
kraleppa: Shan't we check the result of `attach_many`? There is a chance that this GenServer will be killed brutally and previous instance won't be detached

lib/voyager/telemetry/handler.ex
```diff
@@ -0,0 +1,42 @@
...
+  - :export
+  - :logger
+  - :noop
+  """
+
+  use GenServer
+
+  alias Voyager.Telemetry.Events
+
+  @callback handle_event(Events.event(), map(), map(), any()) :: :ok
```
kraleppa, line 8: It's a bit weird that we have a GenServer module that defines behavior. Maybe we shall extract it somehow?

lib/voyager/telemetry/parser.ex
```diff
@@ -0,0 +1,43 @@
...
+
+  @doc "Keeps only metadata used by known event groups."
+  @spec parse_metadata(Events.event(), map()) :: map()
+  def parse_metadata([:phoenix, :live_view | rest], %{socket: socket} = meta) do
+    %{view: inspect(socket.view)}
+    |> maybe_put(:event, meta[:event])
+    |> maybe_exception(rest, meta)
+  end
+
+  def parse_metadata([:voyager | _], meta), do: meta
```
kraleppa: Are we sure that we want to send all metadata? Shan't we whitelist them?

GuzekAlan: These are our custom messages (so we don't need to whitelist them for now)

lib/voyager/telemetry/export.ex
```diff
@@ -0,0 +1,72 @@
...
+          Logger.warning("[telemetry] giving up after #{attempt} retries: #{inspect(reason)}")
+
+          :ok
+        end
+    end
+  end
+
+  defp push(payload) do
+    case Application.get_env(:voyager, :telemetry_push_url) do
+      url when is_binary(url) and url != "" ->
```
hhubert6: Maybe fetching and checking the url should be moved to handler init to not do it with every push?

kraleppa (approved):

test/voyager_web/form_schemas/connection_params_test.exs
```diff
@@ -1,4 +1,5 @@
 defmodule VoyagerWeb.FormSchemas.ConnectionParamsTest do
+  @moduledoc false
```
hhubert6:
```suggestion
```

hhubert6 (changes requested):

hhubert6 (approved):
🐬

srzeszut (approved):

PR #16, opened 2026-05-26
Tests: Add playwright e2e infrastructure

### LINKED ISSUE #7: Add Playwright e2e infrastructure
Prepare full infrastructure for e2e tests in Playwright:
- init Playwright
- create environment for node connection
- create github workflows

srzeszut:


mix.exs
```diff
@@ -64,28 +64,41 @@ defmodule Voyager.MixProject do
     [
       setup: ["deps.get", "ecto.setup", "assets.setup", "assets.build"],
       format: ["format", "cmd npm --prefix assets run format"],
+      "format.all": ["format", "cmd npm --prefix e2e run format"],
```
GuzekAlan: How about adding `e2e.format` and inside `format` calling it (same with `assets.format`) ? Also `precommit` might format all the files

e2e/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
```
GuzekAlan: How about making symlink ?

e2e/playwright.config.ts
```diff
@@ -0,0 +1,57 @@
...
+      testMatch: '**/connect_form.spec.ts',
+    },
+
+    {
+      name: 'connect',
+      use: { ...devices['Desktop Chrome'] },
+      workers: 1,
+      testMatch: '**/connect.spec.ts',
+      dependencies: ['connect-form chromium', 'connect-form firefox'],
+    },
```
GuzekAlan, line 42: Why connect is only for Chrome ?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,93 @@
+name: Playwright Test
+
+on:
+  push:
+    branches:
+      - main
+      - "v*.*"
+  pull_request:
+    branches:
+      - main
```
GuzekAlan, line 10: why not add pull requests to "v*.*" ?

config/runtime.exs
```diff
@@ -4,8 +4,10 @@ if System.get_env("PHX_SERVER") do
   config :voyager, VoyagerWeb.Endpoint, server: true
 end
 
-config :voyager, VoyagerWeb.Endpoint,
-  http: [port: String.to_integer(System.get_env("PORT", "4000"))]
+if config_env() not in [:test, :e2e] do
+  config :voyager, VoyagerWeb.Endpoint,
+    http: [port: String.to_integer(System.get_env("PORT", "4000"))]
+end
```
GuzekAlan: Why not for `:test` ?

config/runtime.exs
GuzekAlan: Vault should be enabled in `:e2e` ?

e2e/playwright.config.ts
srzeszut: This test serves as a connection setup. It separates offline tests from those requiring an active node connection.

e2e/node.sh
hhubert6, line 1: Why `PID_FILE` is created? As far as I can see it's only used for printing `erl` process pid to the terminal, in such case it can be just printed directly without writing to file. Maybe it's worth to consider to use the file to kill that specific process when stopping rather than killing all `erl` processes with `test@127.0.0.1` name.

hhubert6 (commented):
Nice

GuzekAlan (approved):
LGTM 🦈

hhubert6 (approved):
🐬

PR #24, opened 2026-06-02
Feature: create UI for node info

### LINKED ISSUE #20: Create UI for node info view
Continutaion of #3 - the goal is to use the API from #3  and create a UI for it that will be periodically refreshed

kraleppa:
Marking for review: @GuzekAlan @hhubert6 @srzeszut 

I will add tooltips in a separate PR #27


lib/voyager_web/helpers.ex
```diff
@@ -0,0 +1,15 @@
...
+  def ok(state), do: {:ok, state}
+
+  @spec noreply(term()) :: {:noreply, term()}
+  def noreply(state), do: {:noreply, state}
+
+  @spec cont(term()) :: {:cont, term()}
+  def cont(state), do: {:cont, state}
+
+  @spec halt(term()) :: {:halt, term()}
+  def halt(state), do: {:halt, state}
```
GuzekAlan, line 14: Nitpick: `halt/1` is also imported by `Plug.Conn` so wouldn't it conflict?

lib/voyager_web/live/node_info_live.ex
```diff
@@ -1,35 +1,266 @@
...
+  @impl true
+  def handle_info(:refresh, socket) do
+    socket |> fetch_snapshot() |> schedule_refresh() |> noreply()
+  end
+
+  @impl true
+  def handle_async(:snapshot, {:ok, {:ok, snapshot}}, socket) do
+    socket
+    |> assign(:snapshot, AsyncResult.ok(socket.assigns.snapshot, snapshot))
+    |> assign(:last_updated, DateTime.utc_now())
```
GuzekAlan, line 186: Clarification: we want to show `UI` updated or snapshot timestamp ?

kraleppa: Yes, why not?

lib/voyager_web/helpers.ex
kraleppa: Good point - I removed them from modules other than LiveView and LiveComponent so it shouldn't conflict now

lib/voyager_web/formatters.ex
```diff
@@ -0,0 +1,137 @@
...
+
+  Billions use `"B"`, values over ten million use `"M"`; smaller counts are
+  returned as a thousands-separated string with a `nil` unit.
+
+      iex> VoyagerWeb.Formatters.count_parts(2_500_000_000)
+      {2.5, "B"}
+      iex> VoyagerWeb.Formatters.count_parts(1_234)
+      {"1,234", nil}
+  """
+  @spec count_parts(non_neg_integer()) :: {number() | String.t(), String.t() | nil}
```
GuzekAlan:
```suggestion
  @spec count_parts(non_neg_integer()) :: {number(), String.t()} | {String.t(), nil}
```

lib/voyager_web/live/node_info_live.ex
```diff
@@ -1,35 +1,266 @@
...
+              </option>
+            </select>
+          </form>
+          <button
+            type="button"
+            phx-click="refresh"
+            id="refresh-now"
+            title="Refresh now"
+            class="btn btn-sm btn-ghost btn-square"
+          >
```
hhubert6, line 85: Maybe fetching system info is not that expensive but I'd still add here some `phx-throttle`.

lib/voyager_web/formatters.ex
```diff
@@ -0,0 +1,137 @@
...
+    {value, unit} =
+      cond do
+        years > 0 -> {years, "yr #{days}d"}
+        days > 0 -> {days, "d #{hours}h"}
+        hours > 0 -> {hours, "h #{minutes}m"}
+        minutes > 0 -> {minutes, "m"}
+        true -> {seconds, "s"}
+      end
+
+    "#{value}#{unit}"
```
GuzekAlan: Nitpick: `yr #{days}d` as a unit looks weird. Could you just return from `cond` the correct formatted string ?

lib/voyager_web/live/node_info_live.ex
```diff
@@ -1,35 +1,266 @@
...
-  defp fetch_otp_release(node) do
-    # This call should not be here, it's temporary just for showing that Node is connected properly
-    # We will move communication with a node to Node session probably in the future
-    :erpc.call(node, :erlang, :system_info, [:otp_release])
-  catch
-    :error, _ -> nil
+  @impl true
+  def handle_event("refresh", _params, socket) do
+    socket |> fetch_snapshot() |> schedule_refresh() |> noreply()
+  end
```
hhubert6, line 168: Consider canceling previous snapshot fetching task just in case if it's still running to reduce chance of overloading remote node.

lib/voyager_web/live/node_info_live.ex
```diff
@@ -1,35 +1,266 @@
...
   end
 
   @impl true
   def render(assigns) do
     ~H"""
-    <div class="mx-auto max-w-4xl p-8">
-      <h1 class="font-mono text-base-content mb-8 text-2xl font-bold tracking-tight">
-        {@session.node_name}
-      </h1>
+    <div class="max-w-[1536px] mx-auto p-6 sm:p-8">
```
GuzekAlan: Nitpick: Could you use `rem` instead of `px`. It might be good idea to add some `tailiwnd` skills/guidelines to AGENTS.md. What do you think?

lib/voyager_web/live/node_info_live.ex
```diff
@@ -1,35 +1,266 @@
...
+            <NodeInfoComponents.info_card title="Runtime" rows={runtime_rows(snapshot)} />
+
+            <div class="flex h-full min-h-0 flex-col gap-6">
+              <NodeInfoComponents.metric_card
+                title="Schedulers"
+                subtitle="online / total"
+                metrics={[
+                  {"Normal", "#{snapshot.schedulers.online} / #{snapshot.schedulers.total}"},
+                  {"Dirty CPU",
+                   "#{snapshot.schedulers.dirty_cpu_online} / #{snapshot.schedulers.dirty_cpu}"},
```
GuzekAlan, line 144: Nitpick: There are some calculations inside heex template. How about putting it in assigns ?

lib/voyager_web/live/node_info_live.ex
```diff
@@ -1,35 +1,266 @@
...
+    socket
+    |> assign(:refresh_interval, parse_interval(value))
+    |> schedule_refresh()
+    |> noreply()
+  end
+
+  @impl true
+  def handle_info(:refresh, socket) do
+    socket |> fetch_snapshot() |> schedule_refresh() |> noreply()
+  end
```
hhubert6, line 180: In some extreme case scenario when remote node is under high load previous fetch might have not ended and new one is requested. I think it's worth to implement behavior that the interval value specifies after what time new fetch will be requested after previous one ends and returns data. With 5s or higher intervals it will hardly be a case but I think it's a low cost safeguard. What do you think?

lib/voyager_web/live/node_info_live.ex
GuzekAlan, line 1: I don't know if this should go here but could you add some min-width ?

GuzekAlan (commented):
LGTM, only some nitpicks

hhubert6 (commented):
In erlang docs it's explained that ets tables are no longer limited by default by they can be limited with env variable. 
And system info ets limit turns out to be obsolete. Maybe it's worth to change displayed ets limit number to this env variable if it's set.

https://www.erlang.org/doc/apps/erts/erlang.html#system_info/1-system-limits
https://www.erlang.org/doc/apps/stdlib/ets.html#max_ets_tables

lib/voyager_web/live/node_info_live.ex
kraleppa: Tbh, this is more placing strings in correct template than calculations, so I wouldn't add it to assigns

lib/voyager_web/live/node_info_live.ex
kraleppa: Good point, but I'd add it when we make sidebar responsive

GuzekAlan (approved):
LGTM 🦈

kraleppa:
We decided to remove ETS from system limits

Reasons:

- No Hard Limit: In modern BEAM, the number of ETS tables is limited only by available RAM, not a fixed count.
- Metric Mismatch: `ets_count` tracks all active tables (named + anonymous), while `ets_limit` only defines the internal hash registry size for named tables. Comparing them as current / max is technically incorrect.
- Performance vs. Crash: Exceeding this value causes minor CPU lookup overhead (hash collisions) for named tables, never a system crash. Real ETS risk (OOM) is already tracked via memory allocation metrics.

PR #42, opened 2026-06-10
Add fetching Supervision Tree I

### LINKED ISSUE #26: Add fetching Supervision Tree

hhubert6:
cc @kraleppa, @srzeszut, @GuzekAlan


lib/voyager/services/supervision_tree/remote.ex
```diff
@@ -1,26 +1,205 @@
 defmodule Voyager.Services.SupervisionTree.Remote do
   @moduledoc """
-  Currently mock-only. Set `config :voyager, :mock_remote_error, :some_reason`
-  to exercise the LiveView's error branch during development.
+  Thin, safe `:erpc` wrappers for remote node inspection.
+
+  BIF-only: no helper module is loaded on the remote node. Every function
+  returns `{:ok, value} | {:error, reason}` and never lets raw `:erpc`
+  exceptions escape to callers.
```
GuzekAlan: How about removing some of the comments. IMO some of them are explained by code (e. g. ```Calls `:application.which_applications/0` via `:erpc`.```)

lib/voyager/services/supervision_tree/tree_node.ex
```diff
@@ -0,0 +1,47 @@
...
+  @type t :: %__MODULE__{
+          key: String.t(),
+          parent_key: String.t() | nil,
+          pid: pid() | nil,
+          # :name is mostly process's :registered_name if present or it's pid,
+          # but if none present it default to child_id from supervision tree
+          name: term(),
+          type: node_type(),
+          has_children: boolean(),
+          child_count: non_neg_integer(),
```
GuzekAlan: having both these values if they are dependant on each other seems like a place where some mismatch can happen. Do we need both of them? Is it for optimization ?

lib/voyager/services/supervision_tree/walker.ex
```diff
@@ -0,0 +1,489 @@
...
+  defp pid_label(_pid, %{registered_name: name}) when is_atom(name), do: name
+  defp pid_label(pid, _info), do: pid
+
+  # ---------------------------------------------------------------------------
+  # keys & helpers
+  # ---------------------------------------------------------------------------
+
+  defp id_key(pid) when is_pid(pid), do: pid |> :erlang.pid_to_list() |> List.to_string()
+
+  defp ghost_key(parent_key, child_id), do: "#{parent_key}::ghost::#{inspect(child_id)}"
```
GuzekAlan: This one might be problematic for dynamic supervisors since when there is no pid it will produce `undefined` for children

lib/voyager/services/supervision_tree/tree_node.ex
hhubert6: yeah, can be reduced to `child_count` probably

lib/voyager/erpc.ex
```diff
@@ -21,6 +26,10 @@ defmodule Voyager.Erpc do
   @spec call(node(), module(), atom(), [term()]) :: term()
   def call(node, mod, fun, args), do: impl().call(node, mod, fun, args)
 
+  @spec call(node(), module(), atom(), [term()], timeout_time() | call_options()) :: term()
```
srzeszut, line 29:
```suggestion
  @spec call(node(), module(), atom(), [term()], timeout_or_options()) :: term()
```

lib/voyager/services/supervision_tree/walker.ex
```diff
@@ -0,0 +1,499 @@
...
+    errors = root_errors ++ walk_errors ++ hydrate_errors
+
+    if errors == [] do
+      {:ok, result, []}
+    else
+      {:partial, result, errors}
+    end
+  end
+
+  # ---------------------------------------------------------------------------
```
srzeszut, line 98: I'd remove this comments.

lib/voyager/jason_encoders.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Voyager.JasonEncoders do
+  @moduledoc """
+  App-wide `Jason.Encoder` implementations for native Erlang terms that have no
+  default JSON representation.
+
+  These are required so that supervision-tree node values (`name`, `info`) —
+  which stay as raw Erlang terms until push time — can be serialised to the
+  client. Note the implementations are **global**: every `PID`, `Tuple`,
```
kraleppa, line 8: I'm a bit worried that making this global Encoder may somehow impact other parts of the app - especially after I merge the MCP server which uses Jason as well. Maybe we shall scope this serialization for the tree only?

srzeszut (approved):
great job 🐳

kraleppa (approved):

GuzekAlan (approved):
LGTM!

lib/voyager/services/supervision_tree/walker.ex
hhubert6: I think they create a nice separation in this file, so I'd leave if you don't insist

lib/voyager/jason_encoders.ex
hhubert6: All types here have no default implementation so as long as you don't implement other implementation for the same type it should be fine.

PR #45, opened 2026-06-15
Add fetching Supervision Tree II

### LINKED ISSUE #26: Add fetching Supervision Tree

hhubert6:
This pr adds process relationships beyond base supervision tree (links, monitors, monitored by, ports, references).

cc @kraleppa, @srzeszut, @GuzekAlan

### DIFF
--- a/assets/js/hooks/supervision_tree.js
+++ b/assets/js/hooks/supervision_tree.js
@@ -10,6 +10,7 @@ import {
 import { buildStyle, toggleIcon, getColor } from './supervision_tree/styles';
 import {
   elementsFor,
+  relEdgeElement,
   composeLabel,
   edgeId,
   isRealPid,
@@ -17,7 +18,7 @@ import {
 } from './supervision_tree/elements';
 
 /**
- * @import {ServerNode, Info} from './supervision_tree/elements.js'
+ * @import {ServerNode, ServerEdge, Info} from './supervision_tree/elements.js'
  */
 
 cytoscape.use(dagre);
@@ -107,7 +108,8 @@ const SupervisionTree = {
    * @typedef {Object} FullPayload
    * @property {'full'} kind
    * @property {Record<string, ServerNode>} nodes
-
+   * @property {Record<string, ServerEdge>} edges
+   *
    * @typedef {Object} Patch
    * @property {string} name
    * @property {'app'|'supervisor'|'worker'} type
@@ -120,6 +122,8 @@ const SupervisionTree = {
    * @property {Record<string, ServerNode>} added
    * @property {string[]} removed
    * @property {Record<string, Patch>} updated
+   * @property {Record<string, ServerEdge>} edges_added
+   * @property {string[]} edges_removed
    *
    * @param {FullPayload|DeltaPayload} payload
    */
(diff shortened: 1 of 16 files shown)


GuzekAlan (commented):
Monitored by and monitor overlaps each other. Is it possible to detect it and override it ? Or at least add padding to the path so that they are not overlapping ?

Image

assets/js/hooks/supervision_tree.js
```diff
@@ -212,9 +225,29 @@ const SupervisionTree = {
...
+
+      // Relationship edge additions — guard that both endpoints exist (their
+      // rel-node additions were applied earlier in this batch).
+      for (const edge of Object.values(edgesAdded)) {
+        if (this.cy.getElementById(edge.id).nonempty()) continue;
+        const source = this.cy.getElementById(edge.source);
+        const target = this.cy.getElementById(edge.target);
+        if (source.nonempty() && target.nonempty()) {
+          this.cy.add(relEdgeElement(edge));
+        }
```
GuzekAlan, line 253: Are there situations where the source or target are empty or should it be abnormal behaviour? If it is abnormal then some logs might come in handy while debugging later

lib/voyager_web/components/supervision_tree_components.ex
```diff
@@ -185,18 +185,21 @@ defmodule VoyagerWeb.Components.SupervisionTreeComponents do
...
-                  <div class="flex items-center gap-2">
-                    <div class="border-process-monitor w-4 border-b border-dashed" /> Monitor
+                  <div class="gap-[3px] flex items-center">
+                    <span :for={_ <- 1..3} class="w-[4px] h-[2px] bg-process-monitor" />
+                    <span class="ml-1">Monitor</span>
                   </div>
-                  <div class="flex items-center gap-2">
-                    <div class="border-process-monitored-by w-4 border-b border-dashed" />
-                    Monitored by
+                  <div class="gap-[3px] flex items-center">
```
GuzekAlan: Why use `px` and not `rem` ?

GuzekAlan (commented):
When expanding tree with many children (100) the leafs show that there is one child and I can expand but when I do nothing appears:

Image

kraleppa:
Nice job, few nitpicks:
1. I'd disable the effect on drag if possible, since clicking on the edges of the graph does nothing, so I see no use for that
Screenshot 2026-06-29 at 11 19 09
2. I'd consider adding tooltips explaining what given relationship means - it can be done in a separate PR
Screenshot 2026-06-29 at 11 20 00
3. Not sure if this is intended, but showing the path from tree root to selected node does not work in case of other relationships than supervision. It seems quite unintuitive - I'd highlight the path always or never
Screenshot 2026-06-29 at 11 21 04
Screenshot 2026-06-29 at 11 21 10

kraleppa (approved):

assets/js/hooks/supervision_tree.js
hhubert6: I'll add warning.

GuzekAlan (approved):
LGTM! 🦈

srzeszut (approved):
🐳

PR #47, opened 2026-06-18
Feature: add MCP support

### LINKED ISSUE #12: MCP integration
The goal here is to provide an MCP that integrates voyager with coding agents

kraleppa:
This PR introduces MCP support including:
1. MCP component for fetching node state (example for future development)
2. Logic for settings (general settings for the whole app, not only MCP)
3. Architecture around MCP - its own supervision subtree, API for enabling and disabling it

UI with settings page for the MCP will be introduced in the next PR


lib/voyager/mcp/server.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Voyager.MCP.Server do
+  @moduledoc """
+  MCP server exposing BEAM node introspection to coding agents.
+  """
+
+  use Anubis.Server,
+    name: "Voyager",
+    version: "0.1.0",
```
GuzekAlan: how about making a function for `version` to make this and `application.ex` version same ?

GuzekAlan (commented):
Can you check on Cursor CLI since it feels a bit not working 100% correctly

lib/voyager/settings.ex
```diff
@@ -0,0 +1,104 @@
...
+  @spec locked?(atom()) :: boolean()
+  def locked?(key) when is_atom(key) do
+    match?({:ok, _}, Application.fetch_env(:voyager, key))
+  end
+
+  @doc "Returns all persisted settings as a map of `atom key => decoded value`."
+  @spec all() :: %{atom() => term()}
+  def all do
+    Repo.all(from s in Setting, select: {s.key, s.value})
+    |> Map.new(fn {key, encoded} -> {String.to_existing_atom(key), decode(encoded)} end)
```
GuzekAlan: Wouldn't it create problem if we will remove a setting and someone has it sotred in their local db ?

kraleppa: Yeah it would - tbh `all` function is more for debugging and dev-only usage from the iex. I just decided not to parse it to atom

GuzekAlan (approved):
LGTM! 🦈

PR #67, opened 2026-06-26
Improve Supervision Tree controls

### LINKED ISSUE #52: Improve Supervision Tree controls form
- filter out not running apps
- app search
- clear all selected apps button
- store selected apps in query params

hhubert6:
Changed:
- Filtered out not running apps
- Added app search
- Added clear all selected apps button
- Store selected apps and depth in query params
- Added refresh interval select


kraleppa:
The form feels a bit off - as you can see at the very beginning of the video I clicked the `elixir` application and the `iex` was selected. After I unselected iex it switched places with `elixir`.

When I cleared the form, I got like two-staged selection of applications. Firstly, directly when I clicked on application the checkbox changed state to checked, but the component state changed to checked after I clicked on the next application.

https://github.com/user-attachments/assets/dacfc3ea-a9b1-403e-b0f0-47cdb887ea7d

lib/voyager/services/supervision_tree/remote.ex
```diff
@@ -12,13 +12,19 @@ defmodule Voyager.Services.SupervisionTree.Remote do
...
-  @spec list_applications(node()) ::
+  @spec list_running_applications(node()) ::
           {:ok, [{atom(), charlist(), charlist()}]} | {:error, term()}
-  def list_applications(node) do
-    call(node, :application, :which_applications, [], @timeout_fast)
+  def list_running_applications(node) do
+    with {:ok, apps} <- call(node, :application, :which_applications, [], @timeout_fast),
+         {:ok, masters} <- app_masters(node, Enum.map(apps, fn {a, _, _} -> a end)) do
+      running_apps =
+        Enum.zip(apps, masters)
```
kraleppa:
```suggestion
        apps
        |> Enum.zip(masters)
```

GuzekAlan:
Proposal: It might be a good idea to add an indicator on how much apps are selected. You can do it in this PR or create a task for it :)

lib/voyager_web/components/core_components.ex
```diff
@@ -101,6 +101,67 @@ defmodule VoyagerWeb.CoreComponents do
...
+  def interval_select(assigns) do
+    ~H"""
+    <div class="flex items-center gap-2">
+      <label class="font-mono text-base-content/50 tracking-label text-xs uppercase">
+        Auto-refresh
+      </label>
+      <form phx-change="set_interval" id="refresh-interval-form">
+        <select
+          name="interval"
+          id="refresh-interval"
```
kraleppa: If something is in `core_components.ex` it usually means that it can be reused - if so `id` should not be hardcoded

kraleppa (approved):

GuzekAlan (approved):
LGTM! 🦈 

IMO `Clear all` button should be next to search. But i am fine either way

Issue #48, opened 2026-06-18
Add UI for settings of MCP


kraleppa:
The goal is to use already existing API and create to UI for MCP.

It should contain an indicator if MCP is active in the topbar/sidebar, and setting panel that allows changing port, and turning MCP on and off

PR #90, opened 2026-07-06
Refactor: extend settings panel

kraleppa:
For now I moved existing settings to the general purpose settings panel - this is a beginning of #48


srzeszut:
What about adding navbar to connection page and move settings button there?

kraleppa:
> What about adding navbar to connection page and move settings button there?

I tried it - but the navbar feels really empty in that view. Tbh I was thinking about removing navbar entirely and fir everything in the sidebar - we'll consult it with the designer

lib/voyager_web/live/settings_live.ex
```diff
@@ -0,0 +1,83 @@
...
+  alias VoyagerWeb.SettingsLive.DistributionSettings
+
+  @impl true
+  def mount(params, _session, socket) do
+    if connected?(socket) do
+      Phoenix.PubSub.subscribe(Voyager.PubSub, NodeSession.topic())
+    end
+
+    socket
+    |> assign(:return_to, safe_return_to(params["return_to"]))
```
GuzekAlan, line 15: It would be nice to have query params persisted. When user checks applications and chooses depth in SupervisionTree, it will be reset when coming back from settings. What do you think ?

lib/voyager_web/live/settings_live.ex
```diff
@@ -0,0 +1,83 @@
...
+        if String.starts_with?(rest, "/") or String.starts_with?(rest, "\\") do
+          "/"
+        else
+          path
+        end
+
+      _ ->
+        "/"
+    end
+  end
```
GuzekAlan, line 80: Since we don't want malicious content how about checking if path is okay using https://phoenix.hexdocs.pm/Phoenix.Router.html#route_info/4 ?

lib/voyager_web/live/settings_live.ex
```diff
@@ -0,0 +1,83 @@
...
+
+  def handle_info({:distribution_settings, :saved}, socket) do
+    socket
+    |> put_flash(:info, "Distribution suffix saved")
+    |> noreply()
+  end
+
+  def handle_info({:distribution_settings, :locked}, socket) do
+    socket
+    |> put_flash(:error, "Distribution suffix is controlled by application config")
```
GuzekAlan, line 62: It might be a good idea to add an option to put flash from LiveComponent (similar to LiveDebugger). These will be obsolete since they were needed mainly for closing modal. If you think it is outside of the scope of this PR, please let me know

kraleppa: I added this in the #91 - its `push_flash` function

lib/voyager_web/live/settings_live.ex
kraleppa: In the supervision tree as far as I can see checking application does not change query params yet. But I agree when it will be implemented then query params should be persisted

config/dev.exs
```diff
@@ -38,6 +38,3 @@ config :phoenix_live_view,
   debug_heex_annotations: true,
   debug_attributes: true,
   enable_expensive_runtime_checks: true
-
-config :live_debugger, debug_button: false
-config :live_debugger, auto_port: true
```
GuzekAlan, line 43: Could you leave`auto_port` ? `live_debugger` / `live_debugger_tour` are main apps which we might test voyager against and it might make our lives easier

lib/voyager_web/components/shell.ex
```diff
@@ -43,23 +42,23 @@ defmodule VoyagerWeb.Components.Shell do
       </div>
       <div class="navbar-end gap-1">
         <.link
-          navigate={~p"/settings?#{[return_to: settings_return_to(@active_nav, @session)]}"}
+          navigate={
```
GuzekAlan: Nitpick: this redirects outside of `live_session` so warning occurs
> [warning] navigate event to "http://localhost:4000/node/ex%40127.0.0.1" failed because you are redirecting across live_sessions. A full page reload will be performed instead

Maybe use `href` instead ?

GuzekAlan (approved):
Overall looks good to me 🦈

PR #104, opened 2026-07-13
Add SSH connect UI

### LINKED ISSUE #30: Create infrastructure for ssh connections
Create form and db table for ssh connections

srzeszut:


lib/voyager_web/live/connect_live.ex
```diff
@@ -14,140 +11,82 @@ defmodule VoyagerWeb.ConnectLive do
...
-  def handle_event("unpin", %{"id" => id}, socket) do
-    ConnectionActions.unpin(String.to_integer(id))
-    {:noreply, reset_connections(socket)}
-  end
-
-  def handle_event("delete_connection", %{"id" => id}, socket) do
-    ConnectionActions.delete(String.to_integer(id))
-    {:noreply, reset_connections(socket)}
-  end
+  def handle_event(_, _, socket), do: {:noreply, socket}
```
GuzekAlan: What events are discarded here ?

lib/voyager_web/live/connect_live/secret_visibility.ex
```diff
@@ -0,0 +1,43 @@
+defmodule VoyagerWeb.ConnectLive.SecretVisibility do
+  @moduledoc """
+  Shared show/hide behaviour for secret fields (cookies, passwords).
+
+  Attaches a `:handle_event` hook that toggles a per-key `shown?` flag, so
+  LiveComponents rendering `VoyagerWeb.ConnectComponents.secret_field/1` don't
+  need to hand-write a `toggle_x` event for every secret field.
+  """
+
+  use Phoenix.Component
```
GuzekAlan: wouldn't import be enough ?

lib/voyager_web/live/connect_live/recent_connections.ex
```diff
@@ -0,0 +1,120 @@
...
+
+  use Phoenix.Component
+
+  import Phoenix.LiveView
+
+  @type keys :: %{
+          pinned: atom(),
+          recent: atom(),
+          has_pinned: atom(),
+          has_recent: atom()
```
GuzekAlan: `has_pinned` feels a bit weird

lib/voyager_web/live/connect_live/recent_connections.ex
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  @doc "Refetches the connections and re-streams both lists."
+  @spec reset(Phoenix.LiveView.Socket.t()) :: Phoenix.LiveView.Socket.t()
+  def reset(%{assigns: %{recents: %{queries: queries, keys: keys}}} = socket) do
+    {pinned, recent} = Enum.split_with(queries.all(), & &1.pinned)
+
+    socket
+    |> stream(keys.pinned, pinned, reset: true)
+    |> stream(keys.recent, recent, reset: true)
```
GuzekAlan: are streams needed here  (recent/favourite shouldn't be that big chunk of data) ?

lib/voyager_web/live/connect_live/ssh_connect.ex
```diff
@@ -0,0 +1,458 @@
...
+    only: [
+      form_field: 1,
+      secret_field: 1,
+      segmented: 1,
+      name_type_toggle: 1,
+      connect_submit: 1,
+      saved_badge: 1,
+      row_actions: 1,
+      relative_time: 1
+    ]
```
GuzekAlan: wouldn't alias be better here?

lib/voyager_web/live/connect_live/direct_connect.ex
```diff
@@ -0,0 +1,248 @@
...
+    case NodeSession.connect(node_name, cookie, name_type: name_type) do
+      :ok ->
+        cookie_to_store = if remember_cookie, do: cookie, else: nil
+
+        ConnectionActions.upsert_connected(node_name,
+          cookie: cookie_to_store,
+          name_type: name_type
+        )
+
+        {:noreply, redirect(socket, to: ~p"/node/#{node_name}")}
```
GuzekAlan, line 217: `URI.encode` here might be nice

lib/voyager_web/live/connect_live.ex
```diff
@@ -14,140 +11,82 @@ defmodule VoyagerWeb.ConnectLive do
...
+              id="direct-connect"
+              connected?={not is_nil(@connected_session)}
+            />
+
+            <.live_component
+              :if={@mode == :ssh}
+              module={VoyagerWeb.ConnectLive.SshConnect}
+              id="ssh-connect"
+              connected?={not is_nil(@connected_session)}
+            />
```
hhubert6: IMO instead of conditional rendering here if you just hide them via css it won't reset inputs state when switching modes and I think it's a nice touch. As far as I can see it should work right away but if it's not that straight forward then we can leave it as a later improvement.

lib/voyager_web/live/connect_live/ssh_connect.ex
```diff
@@ -0,0 +1,458 @@
...
+        type="button"
+        phx-target={@target}
+        phx-click="fill_ssh_recent"
+        phx-value-id={@conn.id}
+        data-testid="fill-ssh-recent-btn"
+        class="font-mono text-base-content/60 flex min-w-0 flex-1 cursor-pointer items-center gap-2.5 rounded-md px-3 py-2 text-xs transition-colors hover:bg-base-200 hover:text-base-content"
+      >
+        <.icon name="icon-network" class="size-3.5 text-base-content/25 mt-1 shrink-0 self-start" />
+        <div class="flex min-w-0 flex-1 flex-col gap-1">
+          <span class="ml-2 truncate">{@conn.node_name}</span>
```
hhubert6: It looks better aligned to the left
```suggestion
          <span class="ml-2 truncate text-left">{@conn.node_name}</span>
```

lib/voyager_web/live/connect_live/ssh_connect.ex
```diff
@@ -0,0 +1,458 @@
...
+          </div>
+          <div class="flex-[2] min-w-0">
+            <.form_field
+              field={@ssh_form[:ssh_host]}
+              label="SSH Host"
+              placeholder="10.0.0.5"
+              disabled={@connected?}
+            />
+          </div>
+        </div>
```
GuzekAlan, line 86: Why `flex-[1]` `flex-[2]` ? If it can be written without `[]` could you add it

GuzekAlan (commented):
Looks promising 🐙 

Not sure if bug but weird behaviour. When I connect with random string as password it connects (since I have ssh key). This should have some popup, be disabled or there should be some info in connection form

lib/voyager_web/live/connect_live/direct_connect.ex
GuzekAlan: `sigil_p` does it under the hood

hhubert6 (commented):
Nice

srzeszut:
> Looks promising 🐙
> 
> Not sure if bug but weird behaviour. When I connect with random string as password it connects (since I have ssh key). This should have some popup, be disabled or there should be some info in connection form

yes, I am aware of that but I will add this in another pr.

lib/voyager_web/live/connect_live/recent_connections.ex
srzeszut: no, i changed it

lib/voyager_web/live/connect_live/direct_connect.ex
```diff
@@ -1,58 +1,129 @@
...
+      )
+      |> assign(:initialized, true)
+
+    {:ok, socket}
+  end
 
-    ok(socket)
+  def update(assigns, socket) do
+    {:ok, assign(socket, assigns)}
   end
```
GuzekAlan, line 41: Nitpick: assign specific values rather than whatever is passed

lib/voyager_web/live/connect_live/direct_connect.ex
```diff
@@ -1,58 +1,129 @@
...
+  #   """
+  # end
+
   @impl true
   def render(assigns) do
     assigns =
-      assign(
-        assigns,
+      assigns
+      |> assign_new(:id_prefix, fn -> "direct-" end)
```
GuzekAlan: why not put it into mount and use it as module attribute since it is constant value?

lib/voyager_web/live/connect_live/recent_connections.ex
```diff
@@ -0,0 +1,87 @@
...
+    {pinned, recent} = Enum.split_with(queries.all(), & &1.pinned)
+
+    socket
+    |> assign(:pinned_connections, pinned)
+    |> assign(:recent_connections, recent)
+  end
+
+  defp handle_event("pin", %{"id" => id}, socket) do
+    pin_id = String.to_integer(id)
+    socket.assigns.recents.actions.pin.(pin_id)
```
GuzekAlan: I got this error when pinning

```
** (UndefinedFunctionError) function Voyager.Actions.SshConnections.pin/0 is undefined or private
```

piotrsac:
When doing actions:
- Connection type: SSH Tunnel
- Expand advanced
- Change values to a different ones
- Contract advanced
- Click "Connect via SSH" [without a success]

Then the values of the advanced parameters change to a default ones
Seems like a misbehaviour because when the panel is expanded then it doesn't reset

piotrsac:
image

There is a weird box [tooltip?] shown when hovering on Connection type buttons that while being connected to a node [darkened, unclickable buttons]

srzeszut:
> image
> There is a weird box [tooltip?] shown when hovering on Connection type buttons that while being connected to a node [darkened, unclickable buttons]

next pr changes it so i will leave as it is

lib/voyager_web/components/connect_components.ex
```diff
@@ -20,6 +20,44 @@ defmodule VoyagerWeb.ConnectComponents do
     """
   end
 
+  attr :conn, :any, required: true
+
+  attr :disabled, :boolean,
+    default: false,
+    doc: "Disables filling the connect form from this row when a node is already connected"
+
+  def connection_row(assigns) do
```
GuzekAlan: Is this component used anywhere ?

piotrsac:
Here are some designing nits/suggestions
Feel free to not include them or make them a separate, potentially backlog issues as they are all non blocking
(And ofc it's only my opinion)

image

image

the two undertext aren't exactly contradicting each other, because afaik SSH connection underneath also uses BEAM distribution

Also, i feel like the three text elements are repeating themselves a lot without a specific purpose

Screenshot 2026-07-30 at 12 48 52

At the same time im not sure about the font used in "Connect directly over Erlang distribution." - that font we use mostly in less important stuff, usually smaller and on buttons - or bolded like above. Not sure which way to go about it but it seems a little bit inconsistent with the rest of the app
Maybe something like the other font could be a better fit?
 
Screenshot 2026-07-30 at 13 21 18

Both of the selects are "must choose one" and one is dimmed and selected at the same time, and the other is not selected at all. Probably good idea to also check if --name and --sname selection is preserved the same as connection type

Screenshot 2026-07-30 at 13 24 49

lib/voyager_web/components/connect_components.ex
srzeszut: my bad, merge issues

lib/voyager_web/live/connect_live/ssh_connect.ex
```diff
@@ -0,0 +1,473 @@
...
+    do: {:node_name, "Failed to start local Erlang distribution"}
+
+  defp ssh_connect_error({:net_kernel, _}),
+    do: {:node_name, "Failed to start local Erlang distribution"}
+
+  defp ssh_connect_error({:net_kernel_stop, _}),
+    do: {:node_name, "Failed to restart local Erlang distribution"}
+
+  defp ssh_connect_error(:invalid_name_type),
+    do: {:node_name, "Invalid name type — try switching in Advanced settings"}
```
piotrsac: Pretty sure the name type isn't in advanced settings

lib/voyager_web/components/connect_components.ex
```diff
@@ -114,140 +101,234 @@ defmodule VoyagerWeb.ConnectComponents do
...
+      disabled={@disabled}
+      class="btn btn-primary mt-2 w-full gap-2 phx-submit-loading:pointer-events-none phx-submit-loading:opacity-70"
+    >
+      <.icon name={@icon} class="size-4 phx-submit-loading:hidden" />
+      <span class="phx-submit-loading:hidden">{@label}</span>
+      <span class="loading loading-spinner loading-sm hidden phx-submit-loading:inline-flex"></span>
+      <span class="hidden phx-submit-loading:inline">{@loading_label}</span>
+    </button>
     """
   end
```
piotrsac, line 338: It actually never displays "connecting..." label while connecting via SSH

**[WARNING AI EXPLANATION BELOW]**

> What’s wrong
> 
> connect_submit only toggles loading via LiveView’s phx-submit-loading classes:
> Those classes apply only while the submit event is in flight. For SSH, connect_ssh calls start_async and returns immediately:
> After that, phx-submit-loading is gone. @ssh_connecting is true (form faded + button disabled), but the button still shows “Connect via SSH”, not the spinner / “Connecting over SSH…”.
> 
> Direct connect is fine — it’s synchronous, so phx-submit-loading covers the whole wait.
> ```suggestion
> attr :loading, :boolean, default: false
> 
> def connect_submit(assigns) do
>   ~H"""
>   <button
>     type="submit"
>     id={@id}
>     disabled={@disabled}
>     class={[
>       "btn btn-primary mt-2 w-full gap-2 phx-submit-loading:pointer-events-none phx-submit-loading:opacity-70",
>       @loading && "pointer-events-none opacity-70"
>     ]}
>   >
>     <.icon name={@icon} class={["size-4", (@loading && "hidden") || "phx-submit-loading:hidden"]} />
>     <span class={[(@loading && "hidden") || "phx-submit-loading:hidden"]}>{@label}</span>
>     <span class={[
>       "loading loading-spinner loading-sm",
>       if(@loading, do: "inline-flex", else: "hidden phx-submit-loading:inline-flex")
>     ]}></span>
>     <span class={[
>       if(@loading, do: "inline", else: "hidden phx-submit-loading:inline")
>     ]}>{@loading_label}</span>
>   </button>
>   """
> end
> ```
> 
> And in `ssh_connect.ex`
> ```
> <ConnectComponents.connect_submit
>   ...
>   loading={@ssh_connecting}
>   disabled={@connected? or @ssh_connecting}
> />
> ```

rel/app/tauri.sh
```diff
@@ -54,9 +54,20 @@ main() {
...
+      
+      local current_env="${MIX_ENV:-dev}"
+      local ebin="${mix_project_dir}/_build/${current_env}/lib/voyager/ebin"
+      
+      export ELIXIR_ERL_OPTIONS="-epmd_module Elixir.Voyager.ProxyEpmd -pa $ebin"
+      
       cargo_tauri "$@"
       ;;
-    app)
+     app)
```
piotrsac: probably a typo

lib/voyager_web/live/connect_live.ex
```diff
@@ -1,237 +1,111 @@
...
   alias Voyager.NodeSession
-  alias Voyager.Queries.Connections, as: ConnectionQueries
   alias VoyagerWeb.ConnectComponents
-  alias VoyagerWeb.FormSchemas.ConnectionParams
 
   @impl true
   def mount(_params, _session, socket) do
+    if connected?(socket) do
+      Phoenix.PubSub.subscribe(Voyager.PubSub, NodeSession.topic())
+    end
```
piotrsac: I think that connect page subscribes to NodeSesstion.topic() twice in the same process
```suggestion
```

> Router on_mount hook:
> ```
> def on_mount(:observe_node_session, _params, _session, socket) do
>   if connected?(socket) do
>     Phoenix.PubSub.subscribe(Voyager.PubSub, NodeSession.topic())
>   end
> ```
> ConnectLive.mount/3 again:
> ```
> def mount(_params, _session, socket) do
>   if connected?(socket) do
>     Phoenix.PubSub.subscribe(Voyager.PubSub, NodeSession.topic())
>   end
> ```

piotrsac (commented):
Good work, few nits below

lib/voyager_web/live/connect_live.ex
srzeszut: right, merge mistake

lib/voyager_web/live/connect_live/ssh_connect.ex
```diff
@@ -212,8 +207,8 @@ defmodule VoyagerWeb.ConnectLive.SshConnect do
         </:row>
       </RecentConnections.render>
 
-      <p class="font-mono tracking-snug text-base-content/35 mt-6 text-center text-xs">
-        Connects via SSH tunnel
+      <p class="font-mono tracking-snug text-base-content/50 mt-6 text-center text-xs">
```
GuzekAlan, line 210:
```suggestion
      <p class="font-mono tracking-snug text-base-content/70 mt-6 text-center text-xs">
```

GuzekAlan: You could also check other places to go along with  #138

GuzekAlan (approved):
Nice. Good job 🐙

lib/voyager_web/live/connect_live/ssh_connect.ex
hhubert6: I'll wait for this pr to merge and then update #138

PR #107, opened 2026-07-13
Add selecting idividual edge to Supervision Tree

### LINKED ISSUE #64: Add ability to select individual edge in Supervision Tree
Goal is to spotlight individual edge on click to make it easily distinguishable in dense areas and to emphasize edge direction (when has arrow).

piotrsac:


assets/js/hooks/supervision_tree/styles.js
```diff
@@ -193,6 +195,64 @@ export function buildStyle(t) {
...
+    {
+      selector: 'edge.selected',
+      style: {
+        width: 4.5,
+        opacity: 1,
+        'z-index': 20,
+      },
+    },
+    {
+      selector: 'edge[!kind].selected',
```
hhubert6: It seems to work either way but there should be `^` instead of `!` as it matches if attribute is undefined (as it is for regular supervision link).

```suggestion
      selector: 'edge[^kind].selected',
```


I'm adding this attribute for supervision links in other pr (`kind="supervision-link"`) so it will have to be adjusted depending which one will be merged first.

hhubert6 (commented):
Styling of the selection looks good, you could copy it to whole path highlight for consistency and better look (dimming edges not in path as well).

assets/js/hooks/supervision_tree/styles.js
piotrsac: Great catch on the syntax, thank you! I've updated the selector to ^kind. I'll keep an eye on PR so we can adjust it smoothly once one of them merges.

assets/js/hooks/supervision_tree.js
```diff
@@ -74,6 +77,16 @@ const SupervisionTree = {
...
 
+    this.cy.on('mouseover', 'edge', function (event) {
+      event.target.addClass('hover');
+      event.cy.container().style.cursor = 'pointer';
+    });
+
+    this.cy.on('mouseout', 'edge', function (event) {
+      event.target.removeClass('hover');
+      event.cy.container().style.cursor = '';
+    });
```
hhubert6: These `mouseover`/`mouseout` event handlers for `node` and `edge` can be merged

```suggestion
    this.cy.on('mouseover', 'node, edge', function (event) {
      event.target.addClass('hover');
      event.cy.container().style.cursor = 'pointer';
    });

    this.cy.on('mouseout', 'node, edge', function (event) {
      event.target.removeClass('hover');
      event.cy.container().style.cursor = '';
    });
```

assets/js/hooks/supervision_tree.js
```diff
@@ -320,8 +328,12 @@ const SupervisionTree = {
   },
 
   applyHighlight({ path }) {
```
hhubert6: I'd rename it to something like `applyPathHighlight` so it doesn't get confused with edge highlight

hhubert6 (commented):
As you make edges outside the highlighted path dimmed there is a bug when delta payload comes from server and adds some edges, they are added without dimming.

Also I think that if you made edge highlight be reapplied after full and delta payload It's worth doing same for path highlight for consistency. As path highlight is part of selecting individual node then if that node is removed after payload the path highlight should be dropped.

hhubert6 (approved):
Nice 🐬

GuzekAlan (approved):
LGTM 🦈

PR #108, opened 2026-07-14
Add JSON copy modal

### LINKED ISSUE #23: Add modal with JSON snapshot of node info
Add modal with JSON snapshot of node info and copy button to it.

piotrsac:


srzeszut:
image

For me this [icon](https://lucide.dev/icons/file-braces) looks better. What do you think?

lib/voyager_web/live/node_info_live.ex
```diff
@@ -45,6 +48,18 @@ defmodule VoyagerWeb.NodeInfoLive do
     <div class="mx-auto max-w-screen-2xl p-6 sm:p-8">
       <.node_header node_name={@session.node_name} last_updated={@last_updated}>
         <:actions>
+          <button
```
srzeszut: I will add tooltip here. Something like `View snapshot JSON `. What do you think?

piotrsac: Good suggestion! Never thought of that 🚀

lib/voyager_web/components/node_info_components.ex
```diff
@@ -282,6 +282,100 @@ defmodule VoyagerWeb.NodeInfoComponents do
...
+
+  def json_snapshot_modal(assigns) do
+    assigns =
+      assigns
+      |> assign(:title_id, "#{assigns.id}-title")
+      |> assign(:content_id, "#{assigns.id}-content")
+      |> assign(:close_id, "#{assigns.id}-close")
+      |> assign(:copy_id, "#{assigns.id}-copy")
+      |> assign(:backdrop_id, "#{assigns.id}-backdrop")
+      |> assign(:copy_target, "##{assigns.id}-content")
```
GuzekAlan: could you use 1 assign instead of both `content_id` and `copy_target` ?

piotrsac:
> image
> For me this [icon](https://lucide.dev/icons/file-braces) looks better. What do you think?

Good icon choice! In order to make it clearly visible i had to make the button a little bit bigger, as well as refresh button, how does it look to you now?

GuzekAlan (approved):
Only 1 nitpick. Rest looks nice. Good job 🦈

srzeszut (approved):
Great job!!

lib/voyager_web/components/node_info_components.ex
piotrsac: copy_button uses .querySelector(this.el.dataset.copyTarget), so it needs the `#` prefix.
we could use only content_id but in target we would have to pass it like that:
   `target={"##{@content_id}"}`
instead of
   `target={@copy_target}`
and neither option looks ideal to me.

Another way to tackle this problem would be to do something like this:
```js
  const selector = this.el.dataset.copyTarget.startsWith('#') 
    ? this.el.dataset.copyTarget 
    : `#${this.el.dataset.copyTarget}`;
```
(add this to the copy_button)

What do you think?

GuzekAlan: How about moving `#{assigns.id}-content` to constant and then
```elixir
|> assign(:content_id, content_id)
|> assign(:copy_target, "##{content_id}")
```

It will remove some duplication

PR #113, opened 2026-07-20
Add disconnect button to connect page

### LINKED ISSUE #109: Can't easily switch node from connect view
Image

When going back from the node view to the connect page by using "back" on browser (or simply going to the adress manually), there is no other way for changing node than clicking open -> disconnect. You can select other recent or favourite connections (like shown on the screen) but you can't connect to them.

Possible solutions:
- add a button that allows to disconnect from connect page

piotrsac:
## Summary
- Add a disconnect control on the connect page (next to Open) so you can leave a node without opening the node view first
- While a session is active: disable the connect form and favourites/recent fill (UI + server-side `fill_recent` guard); re-stream rows when session state changes so `disabled` stays in sync
- Route connect-page disconnect through `NodeSessionHook` (`:observe_node_session`): same `"disconnect"` → `NodeSession.disconnect/0` → PubSub path as node views; show the shared flash (`Node disconnected` / `Node down`) and stay on `/` while ConnectLive clears `@connected_session`
- Multi-window: shared `NodeSession` loss updates every open connect instance (flash + UI), not only the window that clicked disconnect
- UI polish: Open uses `icon-log-in`; disconnect uses `icon-unplug` (muted error styling)
- Harden Playwright around shared session state (`ensureConnected` / `ensureDisconnected`, serial recent-connections on Chromium + Firefox)


piotrsac:
The changes affect a lot of e2e tests, so its worth checking them out.
At the same time, there is a decision to be made if disconnecting from connect view should display flash messages like the ones @hhubert6 did in #112 either by successful disconnect or node down - or is the connect indicator disappearing enough.
Also im more than happy to hear about visual side of the button itself!

hhubert6 (commented):
> At the same time, there is a decision to be made if disconnecting from connect view should display flash messages like the ones @hhubert6 did in https://github.com/software-mansion/voyager/pull/112 either by successful disconnect or node down - or is the connect indicator disappearing enough.

I think adding a flash as an indicator that something happened especially if it's not an immediate result of a user action in the same view which is the case when there are more than one window of Voyager opened is reasonable. You can add it if you agree.

Image

It kinda looks weird that you have two right arrows with one meaning disconnect and one open inspect. Maybe adding `Disconnect` text and some red color would help 🤔. And maybe replacing `icon-arrow-right` with `icon-log-in`.

assets/css/icons/unplug.svg
hhubert6, line 1: This is now not used anywehere

lib/voyager_web/hooks/node_session_hook.ex
```diff
@@ -32,6 +36,30 @@ defmodule VoyagerWeb.Hooks.NodeSessionHook do
...
+  @doc """
+  Puts the shared disconnect / nodedown flash used across node and connect views.
+  """
+  def put_disconnect_flash(socket, {:node_disconnected, node}) do
+    put_flash(socket, :info, "Node disconnected: #{node}")
+  end
+
+  def put_disconnect_flash(socket, {:nodedown, node}) do
+    put_flash(socket, :error, "Node down: #{node}")
+  end
```
hhubert6: When it's not used anywhere outside the module it should be private. IMO functions on lower level of abstraction should be lower in the file, so I'd move these at the end.

```suggestion
  # Puts the shared disconnect / nodedown flash used across node and connect views.
  defp put_disconnect_flash(socket, {:node_disconnected, node}) do
    put_flash(socket, :info, "Node disconnected: #{node}")
  end

  defp put_disconnect_flash(socket, {:nodedown, node}) do
    put_flash(socket, :error, "Node down: #{node}")
  end
```

hhubert6 (commented):
I feel like it looks better on the right, but maybe you moved it with a reason?

Image

Overall looks solid

piotrsac:
> I feel like it looks better on the right, but maybe you moved it with a reason?
> 
> Image
> Overall looks solid

image
the idea was to match the current-state top bar disconnect which sits close to the node name as well and im biased about where to put it

hhubert6 (approved):
🐬

PR #123, opened 2026-07-22
Enhancement: Add node tooltips in Supervision Tree

hhubert6:


assets/js/hooks/supervision_tree.js
```diff
@@ -519,6 +522,84 @@ const SupervisionTree = {
...
+          return '';
+      }
+    }
+
+    /**
+     * @param {[string, string, number] | undefined} mfa
+     * @param {string} [label]
+     */
+    function parseMfa(mfa, label = '') {
+      if (!Array.isArray(mfa) || mfa.length != 3) return '';
```
piotrsac: small nit
```suggestion
      if (!Array.isArray(mfa) || mfa.length !== 3) return '';
```

assets/js/hooks/supervision_tree.js
piotrsac, line 72: There is no tip repositioning, it should be considered to update the tip location when panning or zooming takes place (although haven't yet found a way to pan while hovering on a node)

piotrsac (commented):
Looks good!
Few suggestions to consider though:
- when for some reason tree collapses, the tip can stay in place even when the node is not visible now
- it can look kinda weird when the node is at the top and the tip covers it, if its possible, maybe the way would be for it to go outside the supervision board, like its done in the legend to the supervision tree
image

- core tooltips use `max-w-xs`; this tip does not. Apparently long Module.fun/arity strings can grow wide and fight the clamp in positionTooltip
- the tooltips are now rendered no matter how small and far the nodes can be, in my opinion it should be done in a way that only after some zooming in, the tip should be shown while hovering
- sometimes the tooltip unreliably renders in wrong position, not sure why and how but it is worth looking for or creating different issue about the tooltips
- we were discussing some e2e test coverage, possibly it could be also one of the tests - if the tip shows, not saying that you should do it right now, but its worth making a note about it somewhere (issue, comment on issue)

lib/voyager_web/components/supervision_tree_components.ex
```diff
@@ -194,6 +194,13 @@ defmodule VoyagerWeb.Components.SupervisionTreeComponents do
             phx-update="ignore"
             class="bg-base-100 relative h-full overflow-hidden rounded-lg"
           >
+            <.tooltip
+              id="supervision-tree-node-snippet"
+              class="absolute"
+              tip_class="max-w-lg! text-sm!"
+            >
+              <:content></:content>
+            </.tooltip>
```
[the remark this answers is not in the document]
hhubert6: https://tailwindcss.com/docs/upgrade-guide#the-important-modifier

assets/js/hooks/supervision_tree/tooltips.js
```diff
@@ -0,0 +1,185 @@
...
+      node.hasClass('hidden') ||
+      this.cy.zoom() < OVERLAY_MIN_ZOOM
+    )
+      return;
+
+    const { name, type, pid, info, app } = node.data();
+
+    const displayName = formatName(info?.registered_name) || formatName(name);
+
+    console.log(pid, escapeHtml(pid));
```
piotrsac: i dont think that console.log is necessary
```suggestion
```

assets/js/hooks/supervision_tree/tooltips.js
```diff
@@ -0,0 +1,185 @@
...
+    this.nodeId = nodeId;
+
+    clearTimeout(this.closeTimer);
+    clearTimeout(this.openTimer);
+
+    this.togglingTooltip = true;
+
+    this.openTimer = setTimeout(() => {
+      this.fillTooltip();
+      setTimeout(() => {
```
piotrsac: Untracked nested `setTimeout`, not cleared after, should consider naming it and clearing the same as `closeTimer` and `openTimer` 
```suggestion
      this.positionTimer = setTimeout(() => {
```

assets/js/hooks/supervision_tree/tooltips.js
```diff
@@ -0,0 +1,185 @@
...
+
+    this.closeTimer = setTimeout(() => {
+      this.toggleTooltipOpen(false);
+      this.togglingTooltip = false;
+    }, TOOLTIP_DELAY_MS);
+  },
+
+  reconcileTooltip() {
+    clearTimeout(this.reconcileTimer);
+    if (this.togglingTooltip) return;
```
piotrsac, line 75: i think we can save some resources by returning if tip is not shown
```suggestion
  reconcileTooltip() {
    if (!this.tooltip?.classList.contains('is-open') || this.togglingTooltip) {
      return;
    }

    clearTimeout(this.reconcileTimer);
```

assets/js/hooks/supervision_tree/tooltips.js
```diff
@@ -0,0 +1,185 @@
...
+
+  initTooltip() {
+    this.tooltip = document.querySelector('#supervision-tree-node-snippet-tip');
+  },
+
+  cleanupTooltip() {
+    clearTimeout(this.openTimer);
+    clearTimeout(this.closeTimer);
+    clearTimeout(this.reconcileTimer);
+  },
```
piotrsac, line 38: I think it lacks closing the tip
```suggestion
  cleanupTooltip() {
    clearTimeout(this.openTimer);
    clearTimeout(this.closeTimer);
    clearTimeout(this.reconcileTimer);
    this.nodeId = null;
    this.toggleTooltipOpen(false);
  },
```
(and if you decide to implement the `positionTimer` it should be also cleared here)

assets/js/hooks/supervision_tree/overlays.js
```diff
@@ -0,0 +1,136 @@
...
+ * mixin (e.g. `this.isCollapsed`, `this.toggleExpandNode`).
+ */
+export const overlayMethods = {
+  /** @type {ReturnType<typeof setTimeout> | undefined} */
+  overlayTimer: undefined,
+  overlays: new Map(),
+
+  initOverlay() {
+    this.overlayLayer = this.el.querySelector('[data-cy-overlays]');
+  },
```
piotrsac, line 19: I think we should create one map per instance of a hook here
```suggestion
  initOverlay() {
    this.overlays = new Map();
    this.overlayLayer = this.el.querySelector('[data-cy-overlays]');
  },
```

lib/voyager_web/components/supervision_tree_components.ex
```diff
@@ -194,6 +194,14 @@ defmodule VoyagerWeb.Components.SupervisionTreeComponents do
...
             class="bg-base-100 relative h-full overflow-hidden rounded-lg"
           >
+            <.tooltip
+              id="supervision-tree-node-snippet"
+              class="absolute"
+              tip_class="max-w-lg! text-sm!"
+            >
+              <span class="hidden"></span>
+              <:content></:content>
+            </.tooltip>
```
piotrsac: Not totally sure this is worth changing, but I wonder if a bare `<.portal> `tip might fit better here than `<.tooltip>`? The Tooltip hook on an empty trigger feels a bit unused - we mainly seem to need the portaled tip element
```suggestion
            <.portal id="supervision-tree-node-snippet-portal" target="#tooltip-portal-root">
              <div
                id="supervision-tree-node-snippet-tip"
                role="tooltip"
                phx-update="ignore"
                class={[
                  "tooltip-pop bg-base-100 text-base-content rounded-box max-w-lg px-3 py-2",
                  "ring-base-content/15 text-sm leading-relaxed shadow-lg ring-1"
                ]}
              >
              </div>
            </.portal>
```

assets/js/hooks/supervision_tree/tooltips.js
```diff
@@ -0,0 +1,185 @@
...
+    if (!this.nodeId) return;
+
+    const node = this.cy.getElementById(this.nodeId);
+    if (
+      node.empty() ||
+      node.hasClass('hidden') ||
+      this.cy.zoom() < OVERLAY_MIN_ZOOM
+    )
+      return;
+
```
piotrsac: I think with an early exit here we should close the tooltip before returning
```suggestion
    ) {
      this.toggleTooltipOpen(false);
      return;
    }

```

piotrsac (commented):
It looks really good now, good idea with mixin split! Wouldn't be me though if i haven't pick on something. A few blockers around tip leftover log / per-instance state; rest are nits.

assets/js/hooks/supervision_tree/tooltips.js
hhubert6: The way it works now checking against `this.nodeId` later cuts off unnecessary calculations but enables behavior like hiding tooltip when zoomed out and showing again when zoomed in without any mouse movement. The idea is that when `this.nodeId` is set and conditions are met (e.g. node is visible and proper zoom level) tooltip should be shown.

lib/voyager_web/components/supervision_tree_components.ex
hhubert6: Makes sense

assets/js/hooks/supervision_tree/tooltips.js
```diff
@@ -0,0 +1,188 @@
...
+ * shares `this.cy` with the rest of the hook.
+ */
+export const tooltipMethods = {
+  /** @type {ReturnType<typeof setTimeout> | undefined} */
+  closeTimer: undefined,
+  /** @type {ReturnType<typeof setTimeout> | undefined} */
+  openTimer: undefined,
+  /** @type {ReturnType<typeof setTimeout> | undefined} */
+  reconcileTimer: undefined,
+  /** @type {boolean} */ togglingTooltip: false,
```
piotrsac, line 24:
```suggestion
  /** @type {ReturnType<typeof setTimeout> | undefined} */
  reconcileTimer: undefined,
  /** @type {ReturnType<typeof setTimeout> | undefined} */
  positionTimer: undefined,
  /** @type {boolean} */ togglingTooltip: false,
```

assets/js/hooks/supervision_tree/tooltips.js
```diff
@@ -0,0 +1,188 @@
...
+    clearTimeout(this.positionTimer);
+    this.nodeId = null;
+    this.toggleTooltipOpen(false);
+  },
+
+  scheduleShowTooltip(nodeId) {
+    this.nodeId = nodeId;
+
+    clearTimeout(this.closeTimer);
+    clearTimeout(this.openTimer);
```
piotrsac, line 44:
```suggestion
    clearTimeout(this.openTimer);
    clearTimeout(this.positionTimer);
```

assets/js/hooks/supervision_tree/tooltips.js
```diff
@@ -0,0 +1,188 @@
...
+        this.togglingTooltip = false;
+      });
+    }, TOOLTIP_DELAY_MS);
+  },
+
+  scheduleCloseTooltip() {
+    this.nodeId = null;
+
+    clearTimeout(this.closeTimer);
+    clearTimeout(this.openTimer);
```
piotrsac, line 62:
```suggestion
    clearTimeout(this.openTimer);
    clearTimeout(this.positionTimer);
```

piotrsac (approved):
🦭

PR #124, opened 2026-07-23
Disable ssh connections if proxy_epmd is not active

srzeszut:


lib/voyager/proxy_epmd.ex
```diff
@@ -56,6 +56,10 @@ defmodule Voyager.ProxyEpmd do
     end
   end
 
+  def active? do
```
kraleppa, line 60: You're missing spec

lib/voyager/proxy_epmd.ex
```diff
@@ -56,6 +56,10 @@ defmodule Voyager.ProxyEpmd do
...
+    :persistent_term.get(:voyager_epmd_module) == __MODULE__
```
kraleppa: If the key does not exist it will raise an error

kraleppa (approved):

PR #125, opened 2026-07-24
Improvement: remove footer and move MCP and connections status

kraleppa:
Screenshot 2026-07-24 at 12 42 34


srzeszut:
After shrinking settings icon overlaps logo and connected indicator.
image

srzeszut:
image
I am wondering if the name is long, maybe we can remove the 'Connected' caption.
What do you think?
Or maybe don't truncate the name if there's plenty of space?

srzeszut:
I will move disconnect button outside this connected section and maybe add `Disconnect` label beside. What do you think?

piotrsac:
> image I am wondering if the name is long, maybe we can remove the 'Connected' caption. What do you think? Or maybe don't truncate the name if there's plenty of space?

Agree with that, usually there is plenty of space and `max-w-40` on session node name doesn't seem big enough even for normal-lengthed node names

lib/voyager_web/components/shell.ex
```diff
@@ -116,24 +109,34 @@ defmodule VoyagerWeb.Components.Shell do
 
   defp mcp_status_indicator(assigns) do
     ~H"""
-    <.tooltip id={"mcp-status-tip-#{@status.alive?}"} interactive position="bottom">
+    <.tooltip
+      id={"mcp-status-tip-#{@status.alive?}-#{:erlang.phash2(@status.url)}"}
+      interactive
+      position="right"
```
piotrsac: Image
I would consider displaying the tooltip on top, especially while the sidebar is expanded

```suggestion
      position="top"
```

piotrsac:
In #113 im also working on a disconnect button and the conclusion was to make it in error (red) color but with lower opacity to make it not stand out as much, yet for it to be easily seen

piotrsac:
image
image

Green pulsing icons seem to be different in MCP and connection, is that intentional?

piotrsac:
Looks good, just a few finishing nits

image

The MCP block sits outside the menu with ml-1.5 p-3, while nav items use menu padding + margin-inline: auto in compact mode, i think they should be aligned

lib/voyager_web/components/shell.ex
```diff
@@ -319,16 +333,42 @@ defmodule VoyagerWeb.Components.Shell do
...
+      >
         {@session.node_name}
-      </div>
-      <div class="flex-1"></div>
-      <div>v0.1.0</div>
-    </footer>
+      </span>
+      <.tooltip id="disconnect-tip" position="bottom">
+        <button
+          type="button"
```
piotrsac, line 377: just a nit for consistency and a11y
```suggestion
          id="disconnect"
          type="button"
          title="Disconnect"
          aria-label="Disconnect"
```

piotrsac (approved):
🦭

srzeszut:
I wonder if this tooltip can be similar to help tooltips in node info. Instead of external icon there can be copy button. What do you think? 
image
image

srzeszut (approved):

PR #130, opened 2026-07-27
Unify icons/buttons

piotrsac:
Unify square icon-button sizes behind CSS tokens so chrome controls can be tuned in one place.

- Add toolbar-btn / toolbar-icon and dense toolbar-btn-sm / toolbar-icon-sm (tokens + utilities)
- Migrate settings, sidebar, refresh, zoom, secret toggles, disconnect/pin/remove, details panel, etc. onto those classes
- Drive compact sidebar width and menu padding from the same tokens so icons stay centered when pad/size change
- Move disconnect outside the Connected chip (still to its right)
- Keep close-panel X as an optical exception (size-6) because the glyph reads smaller than refresh at token size


piotrsac:
@hhubert6 probably mostly visual review of the buttons needed

lib/voyager_web/components/connect_components.ex
```diff
@@ -85,16 +88,16 @@ defmodule VoyagerWeb.ConnectComponents do
...
-            <.icon name="icon-power" class="size-3.5" />
+            <.icon name="icon-power" class="toolbar-icon-sm" />
           </button>
         </div>
         <.link
           href={~p"/node/#{@session.node_name}"}
           class="btn btn-success btn-xs ml-3 shrink-0 gap-1"
         >
-          Open <.icon name="icon-log-in" class="size-3.5" />
+          Open <.icon name="icon-log-in" class="size-4" />
```
hhubert6: Why not `toolbar-icon-sm`?

hhubert6 (approved):
Looking good 🐬

PR #131, opened 2026-07-28
Bug: fix sidebar toggling

kraleppa:


lib/voyager_web/utils/url.ex
```diff
@@ -0,0 +1,76 @@
+defmodule VoyagerWeb.Utils.URL do
```
GuzekAlan, line 1: If this is from LiveDebugger then could you also copy related unit tests for this module ?

lib/voyager_web/components/shell.ex
```diff
@@ -335,8 +312,29 @@ defmodule VoyagerWeb.Components.Shell do
...
+  # navigation to another page.
+  defp nav_path(path, mode) when mode in ["compact", "full"],
+    do: URL.put_query_param(path, "sidebar", mode)
+
+  defp nav_path(path, _mode), do: path
+
+  defp toggle_sidebar_path(current_url, sidebar_mode) do
+    next_mode = if sidebar_mode == "compact", do: "full", else: "compact"
+    URL.put_query_param(current_url || "/", "sidebar", next_mode)
+  end
```
GuzekAlan, line 357: It does not follow the current screen width. There is an issue when there is no `sidebar` in URL and on small width (when sidebar is compact) toggle sets `compact` (user doesn't see change)

GuzekAlan (approved):
🐙

PR #133, opened 2026-07-28
Change DaisyUI input caret into custom one

piotrsac:
DaisyUI paints the <select> caret as two 4×4px CSS linear-gradient
   triangles joined at a fractional 16.1px offset. At non-100% zoom (and when
   the control itself lands on subpixel coordinates) those halves misalign —
   the caret looks split, deformed, or half-missing.

   Kill DaisyUI's dual-gradient caret globally, and render a single SVG caret
   via `.select-caret` wrapper so there is no seam to break under fractional
   device pixels. Color follows currentColor.

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -3,6 +3,7 @@
 @import './styles/tokens.css';
 @import './styles/reset.css';
 @import './styles/number-input.css';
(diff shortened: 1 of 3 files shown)


kraleppa (approved):
🦈

PR #135, opened 2026-07-29
Feature: add ToU popup, telemetry switch and install id

kraleppa:


lib/voyager_web/live/settings_live/telemetry_settings.ex
```diff
@@ -0,0 +1,80 @@
...
+        |> assign(:enabled?, new_value)
+        |> noreply()
+
+      {:error, _} ->
+        socket
+        |> push_flash(:error, "Failed to toggle telemetry")
+        |> assign(:toggle_revision, socket.assigns.toggle_revision + 1)
+        |> noreply()
+    end
+  end
```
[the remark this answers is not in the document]
kraleppa: The toggle is already rendered disabled when the setting is controlled by application config, so this event cannot normally occur.

lib/voyager_web/router.ex
```diff
@@ -19,7 +19,10 @@ defmodule VoyagerWeb.Router do
 
     live_session :connect,
       layout: {VoyagerWeb.Layouts, :connect},
-      on_mount: [{VoyagerWeb.Hooks.NodeSessionHook, :observe_node_session}] do
+      on_mount: [
+        {VoyagerWeb.Hooks.NodeSessionHook, :observe_node_session},
+        VoyagerWeb.Hooks.OnboardingHook
+      ] do
```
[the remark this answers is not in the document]
kraleppa: this is intentional. The popup links to Settings so users can review or disable telemetry before continuing. Showing the same blocking popup on /settings would make that link unusable.

lib/voyager/telemetry/handler/export.ex
```diff
@@ -23,6 +25,7 @@ defmodule Voyager.Telemetry.Handler.Export do
     case config do
       %{push_url: push_url, api_key: api_key} when is_binary(push_url) and is_binary(api_key) ->
         payload = %{
+          install_id: InstallId.get(),
```
GuzekAlan: This field will be dropped by ingest so use `metadata` to pass this information in telemetry logs

lib/voyager_web/live/settings_live/telemetry_settings.ex
```diff
@@ -0,0 +1,80 @@
...
+  alias Voyager.Settings
+  alias Voyager.Telemetry
+
+  @impl true
+  def update(assigns, socket) do
+    socket
+    |> assign(assigns)
+    |> assign_new(:locked?, fn -> Settings.locked?(:telemetry_enabled) end)
+    |> assign_new(:enabled?, fn -> Settings.get(:telemetry_enabled, true) end)
+    |> assign_new(:toggle_revision, fn -> 0 end)
```
GuzekAlan: Nitpick: why not use `mount` for static assign once assigns. Also when assigning assigns you could pass explicit assigns so that they are filtered

lib/voyager_web/components/layouts.ex
```diff
@@ -73,6 +75,76 @@ defmodule VoyagerWeb.Layouts do
...
+      :if={@show}
+      id="onboarding-modal"
+      role="dialog"
+      aria-modal="true"
+      aria-labelledby="onboarding-modal-title"
+      class="modal modal-open"
+    >
+      <div class="modal-box border-base-300 max-w-xl border shadow-2xl">
+        <div class="mb-4 flex items-center gap-3">
+          <div class="bg-primary/10 text-primary rounded-box size-11 flex shrink-0 items-center justify-center">
```
GuzekAlan: Design suggestion: it might be good idea to use `bg-info` and `text-info-content`. It now doesn't look good now but after changing these colors it might be good. 

But I am fine either way

lib/voyager/telemetry/manager.ex
```diff
@@ -26,14 +29,19 @@ defmodule Voyager.Telemetry.Manager do
     handler_module = handler_module(opts)
     handler_config = handler_config(opts)
 
+    # Cached in :persistent_term (rather than read from the DB on every
+    # dispatch) since `handle_event/4` runs synchronously in the caller's
+    # process for every telemetry event.
+    :persistent_term.put(@enabled_key, Settings.get(:telemetry_enabled, true))
```
piotrsac: im a little bit concerned that when someone launches an app, it afaik starts sending telemetry right away, because `:telemetry_enabled` is true by default. That means that the telemetry could be sent even if someone is stuck at the entry screen with a pop-up and didn't click continue, meaning he hasn't explicitly accepted the ToU. ofc the telemetry sent then isn't really invasive in any way, but i would think about enabling the telemerry after clicking continue (but we would lose some data).
We already have `:terms_accepted` so we can use that or consider updating the popup message so it more explicitly tells that the telemetry starts now, and not after clicking "continue"

lib/voyager_web/hooks/onboarding_hook.ex
```diff
@@ -0,0 +1,27 @@
...
+      |> assign(:show_onboarding?, not Settings.get(:terms_accepted, false))
+      |> attach_hook(:dismiss_onboarding, :handle_event, &handle_dismiss/3)
+
+    {:cont, socket}
+  end
+
+  defp handle_dismiss("dismiss-onboarding", _params, socket) do
+    Settings.put(:terms_accepted, true)
+    {:halt, assign(socket, :show_onboarding?, false)}
+  end
```
piotrsac, line 30: I think that there should be unhappy put path handled
```suggestion
  defp handle_dismiss("dismiss-onboarding", _params, socket) do
    case Settings.put(:terms_accepted, true) do
      {:ok, _} ->
        {:halt, assign(socket, :show_onboarding?, false)}
      {:error, _} ->
        {:halt, put_flash(socket, :error, "Failed to save terms acceptance")}
    end
  end
```

config/config.exs
```diff
@@ -4,6 +4,8 @@ config :voyager,
   ecto_repos: [Voyager.Repo],
   generators: [timestamp_type: :utc_datetime]
 
+config :voyager, :terms_of_use_url, "https://swmansion.com/legal/voayager/terms-of-use/"
```
srzeszut:
```suggestion
config :voyager, :terms_of_use_url, "https://swmansion.com/legal/voyager/terms-of-use/"
```

lib/voyager/telemetry/install_id.ex
```diff
@@ -0,0 +1,56 @@
...
+
+  alias Voyager.Settings
+
+  require Logger
+
+  @cache_key {__MODULE__, :install_id}
+  @setting_key :install_id
+
+  @doc """
+  Returns the installation id, generating and persisting one on first call.
```
srzeszut, line 21: why not generating and persisting on telemetry init and than use here value from `:persistent_term`?

test/voyager_web/live/connect_live_test.exs
```diff
@@ -88,6 +89,28 @@ defmodule VoyagerWeb.ConnectLiveTest do
     end
   end
 
+  describe "onboarding popup" do
+    test "shows on first launch and dismissing it persists acceptance", %{conn: conn} do
```
piotrsac, line 96: I think i would add `config :voyager, :terms_accepted, true` to `test.exs` for most tests to work without a modal (like on e2e tests), and here i would put 
```suggestion
    setup do
      # test.exs locks :terms_accepted so unrelated LiveViews skip the modal.
      # Clear it here so we exercise the real first-launch / DB-backed path.
      Application.delete_env(:voyager, :terms_accepted)
      on_exit(fn -> Application.put_env(:voyager, :terms_accepted, true) end)
      :ok
    end

    test "shows on first launch and dismissing it persists acceptance", %{conn: conn} do
```
It works nevertheless though

piotrsac:
Grok yells at me that there is a test coverage gap: i'll leave that for you to decide if it's justified and is it for this or other PR/issue



> Not a runtime bug — a **coverage gap**. The feature claim is: every exported event carries an anonymous install id in `metadata`. That wiring lives only here:
> 
> ```28:33:lib/voyager/telemetry/handler/export.ex
>         payload = %{
>           event: Parser.parse_event(event),
>           measurements: Parser.parse_measurements(event, measurements),
>           metadata: export_metadata(event, metadata),
>           ts: System.system_time(:millisecond)
>         }
> ```
> 
> ```92:96:lib/voyager/telemetry/handler/export.ex
>   defp export_metadata(event, metadata) do
>     event
>     |> Parser.parse_metadata(metadata)
>     |> Map.put(:install_id, InstallId.get())
>   end
> ```
> 
> Tests cover the pieces separately:
> 
> | Covered | Not covered |
> |---|---|
> | `InstallId.get/0` generate / persist / cache | `Export.handle_event/4` builds payload with `install_id` |
> | Manager enable/disable forwarding | POST body actually contains that field |
> | Parser in isolation | Integration: Manager → Export → metadata |
> 
> There is **no** `export_test.exs` (or similar) under `test/voyager/telemetry/`.
> 
> ### Why it matters
> 
> Someone can break the feature without a red CI:
> 
> - Delete or “refactor away” `Map.put(:install_id, …)`
> - Put `install_id` on the wrong key / nesting
> - Call `InstallId.get()` only in a code path that never runs
> - Parser output shape changes and overwrites / drops the field later
> 
> `InstallId` tests would still pass. Manager tests would still pass. Ingest would just stop getting correlatable install ids until someone notices in production.
> 
> Severity: **nice-to-have / should-fix for a feature PR**, not a user-facing defect.

lib/voyager/telemetry/manager.ex
kraleppa: true, I changed that

lib/voyager/telemetry/install_id.ex
```diff
@@ -0,0 +1,56 @@
...
+    case Settings.put(@setting_key, install_id) do
+      {:ok, _setting} ->
+        :ok
+
+      {:error, reason} ->
+        Logger.warning("Failed to persist telemetry install id: #{inspect(reason)}")
+    end
+
+    install_id
+  end
```
piotrsac, line 58: A failed first write still caches the id in :persistent_term, so it won’t retry and the id is lost on restart. Prefer not caching until persist succeeds, or retry on later get/0.

piotrsac (approved):
Nice work! 🦭

PR #142, opened 2026-07-31
Change default theme to auto and use OS theme

piotrsac:


GuzekAlan (approved):
🐙

rel/app/src-tauri/src/lib.rs
```diff
@@ -74,10 +74,18 @@ pub fn run() {
 fn create_window(app_handle: &tauri::AppHandle, port: u16) {
     let n = app_handle.webview_windows().len() + 1;
     let url = tauri::WebviewUrl::External(format!("http://127.0.0.1:{port}").parse().unwrap());
+    // Inject the OS appearance before page scripts run so Auto mode can paint
+    // the correct DaisyUI theme on first frame (webview prefers-color-scheme
+    // often disagrees with the real OS theme).
+    let theme_init = format!(
+        r#"window.__VOYAGER_OS_THEME__="{}";"#,
+        utils::os_theme_hint()
+    );
```
hhubert6: This script hardcodes os theme detected when app is started and is run on every app navigation so in automode it overrides `onThemeChanged` change and uses start up value and not current system value.

lib/voyager_web/components/layouts/root.html.heex
```diff
@@ -28,43 +28,123 @@
...
+        };
+
+        const listenForOsThemeChanges = async () => {
+          const win = window.__TAURI__?.window?.getCurrentWindow?.();
+          if (!win?.onThemeChanged) return false;
+
+          try {
+            await win.onThemeChanged(({ payload }) => {
+              if (isTheme(payload)) applyOsTheme(payload);
+            });
```
hhubert6: The docs suggest to `unlisten` this event listener when going out of scope (https://v2.tauri.app/reference/javascript/api/namespacewebviewwindow/#onthemechanged). And now it adds this listener on every navigation so it might be a good idea.

hhubert6 (approved):
Nice 🐬

PR #144, opened 2026-08-03
Chore: bump dependencies

GuzekAlan:


lib/voyager/mcp/router.ex
```diff
@@ -16,11 +16,18 @@ defmodule Voyager.MCP.Router do
   plug :match
   plug :dispatch
 
+  # Anubis passes `:subscriber_metadata` to the `forward` without using MFA format which results in errors.
```
GuzekAlan, line 19: I've added issue to AnubisMCP regarding this problem so we can change it if they update their code

kraleppa (approved):

PR #151, opened 2026-08-04
Fix flaky e2e tests

hhubert6:
Sample workflow runs: https://github.com/software-mansion/voyager/actions/workflows/bench-playwright-flakiness.yml

### DIFF
--- a/.github/workflows/bench-playwright-flakiness.yml
+++ b/.github/workflows/bench-playwright-flakiness.yml
@@ -32,7 +32,7 @@ concurrency:
 
 jobs:
   benchmark:
-    name: Benchmark E2E tests flakiness
+    name: Benchmark
     runs-on: ubuntu-24.04
 
     env:
--- a/.github/workflows/bench-playwright-flakiness.yml
+++ b/.github/workflows/bench-playwright-flakiness.yml
@@ -1,14 +1,27 @@
-name: Playwright Test
+name: Benchmark E2E Flakiness
 
 on:
-  push:
-    branches:
-      - main
-      - "v*.*"
+  ################################################
+  # TODO: Remove this trigger
+  # temporary trigger to index workflow on github
+  ################################################
   pull_request:
     branches:
       - main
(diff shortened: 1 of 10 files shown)


e2e/aggregate-flakiness.mjs
```diff
@@ -0,0 +1,291 @@
...
+
+let dir;
+let write = false;
+let threshold = null;
+for (const a of process.argv.slice(2)) {
+  if (a === '--write') write = true;
+  else if (a.startsWith('--threshold='))
+    threshold = Number(a.slice('--threshold='.length));
+  else if (!a.startsWith('--')) dir = a;
+}
```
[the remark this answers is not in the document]
hhubert6: Whole workflow is quite expansive to run so instead of aggressive exit with wrong input (which is already optional) it's better to silently skip this argument.

e2e/playwright.config.ts
```diff
@@ -43,7 +43,7 @@ export default defineConfig({
     },
 
     {
-      name: 'recent-connections',
+      name: 'recent-connections chrome',
```
GuzekAlan: Nitpick: Everywhere else there is `chormium`. Can you unify the name ?

.github/workflows/bench-playwright-flakiness.yml
```diff
@@ -0,0 +1,106 @@
...
+          cache: "npm"
+          cache-dependency-path: "./e2e/package-lock.json"
+
+      - name: Install Node dependencies
+        working-directory: ./e2e
+        run: npm ci
+
+      - name: Check prettier
+        working-directory: ./e2e
+        run: npx prettier . --check
```
GuzekAlan: Is prettier needed ?

GuzekAlan (commented):
Looks good to me only small nitpicks

Great Job

🐙

GuzekAlan (approved):
LGTM! 
🐙

PR #157, opened 2026-08-06
Fix dependabot cargo config

piotrsac:
Switch Tauri elixirkit from a Mix deps/ path to a git dep (same SHA as mix.lock) so Dependabot cargo works. Add dev/elixirkit_check_rev.sh (CI + precommit) to keep Mix and Cargo revs in sync.


GuzekAlan (approved):
Nice 🐙

GuzekAlan:
Please wait for copilot review

PR #158, opened 2026-08-06
Add tooltip for disconnect_from_connect and remove redundant title tips

piotrsac:
Add tooltip and remove titles as they were redundant and covering each other


srzeszut (approved):
🐋

PR #161, opened 2026-08-07
Add copy button confirmation icon change

piotrsac:
- Give icon-only copy buttons visible feedback by briefly swapping the icon to a check on success or an x on failure, then restoring the copy icon.
- Add a Lucide check icon and keep existing label/aria feedback for labeled copy buttons.


kraleppa (approved):
🦈

PR #176, opened 2026-08-17
Bug: Persist MCP enabled setting in the db

kraleppa:


lib/voyager/mcp/endpoint_manager.ex
hhubert6, line 13: `:mcp_enabled` is now configurable as well so it should be mentioned

hhubert6 (commented):
IMO as the settings can be overwritten by config it should be handled the same as `telemetry_enabled` with toggle disabled when setting is `locked?` and handling toggle in endpoint manager whether saving in settings succeeded or not.

lib/voyager/mcp/endpoint_manager.ex
hhubert6: In these branch setting in db stays as `mcp_enabled: false` but the gen server becomes `enabled: true`

kraleppa:
@claude review opus

hhubert6 (approved):
🐬

PR #181, opened 2026-08-18
Fix: Ensure epmd is started before connections

srzeszut:


lib/voyager/services/distribution.ex
```diff
@@ -74,23 +72,62 @@ defmodule Voyager.Services.Distribution do
...
+    else
+      Logger.warning(
+        "net_kernel.start/2 failed and epmd appears down. Starting epmd and retrying..."
+      )
+
+      restart_epmd_and_distribute(name_type, reason)
+    end
+  end
+
+  defp restart_epmd_and_distribute(name_type, reason) do
```
piotrsac: restart is a misleading name here, `Daemon.start()` does not restart

piotrsac:
wdyt about starting epmd at boot of the app also, just so it could be potentially ready by the time someone tries to connect?

piotrsac:
@claude review opus

srzeszut:
> wdyt about starting epmd at boot of the app also, just so it could be potentially ready by the time someone tries to connect?

forgot to push that, thanks

srzeszut:
@claude review opus

piotrsac:
@srzeszut 
Note that this issue couldn't be reproduced at this moment - seems like some change while it was frozen fixed that issue - now the ssh connection works fine without `epmd` running

Closing the issue without further investigation, it can be re-opened if the issue comes back
It is advised to leave the branch `ensure-epmd-is-started-before-connection` without deleting it, for potential re-opening of the issue

PR #184, opened 2026-08-20
Toggle setting UI for pids

GuzekAlan:


assets/js/hooks/supervision_tree/elements.js
```diff
@@ -105,28 +106,39 @@ export function initialIsCollapsedState({ child_count, children_keys }) {
...
-  if (typeof name === 'string') return name;
+  if (Array.isArray(name))
+    return name.map((part) => formatName(part, format)).join(':');
+  if (typeof name === 'string') {
+    return isRealPid(name) ? formatPid(name, format) : name;
+  }
   return String(name);
 }
 
+export function formatPid(pid, format = 'distribution') {
```
piotrsac: If im not mistaken its the same logic that is in `maybe_localize_pid` in Elixir in `lib/voyager_web/formatters.ex` on lines 161-165 - thats duplicated code‎

lib/voyager_web/live/settings_live/pid_format_settings.ex
```diff
@@ -0,0 +1,138 @@
...
+  end
+
+  @impl true
+  def render(assigns) do
+    data = [
+      {"icon-network", "Distribution", "<123.23.423>",
+       "Keeps the remote node index, which identifies a process across a cluster."},
+      {"icon-laptop", "Local", "<0.23.423>",
+       "Replaces the node index with <span class='font-mono'>0</span>, so you can use it on remote shells."}
+    ]
```
piotrsac: perhaps we could use a little bit more descriptive name than just `data`

lib/voyager_web/live/settings_live/pid_format_settings.ex
```diff
@@ -0,0 +1,138 @@
...
+    put_format(socket, :local)
+  end
+
+  defp put_format(socket, format) do
+    cond do
+      socket.assigns.locked? ->
+        noreply(socket)
+
+      socket.assigns.pid_format == format ->
+        noreply(socket)
```
piotrsac: please change for `{:noreply, socket}` to ensure consistency across project

test/voyager_web/live/supervision_tree_live/details_panel_test.exs
```diff
@@ -93,6 +95,26 @@ defmodule VoyagerWeb.SupervisionTreeLive.DetailsPanelTest do
...
+
+      html =
+        render_component(&DetailsPanelComponents.node_label/1, %{
+          panel_id: "details-panel",
+          node: node,
+          pid_format: :local
+        })
+
+      assert html =~ "&lt;0.45.0&gt;"
+      refute html =~ "&lt;123.45.0&gt;"
```
piotrsac: quote from agents.md
``` 
- **Never** tests again raw HTML, **always** use `element/2`, `has_element/2`, and similar: `assert has_element?(view, "#my-form")`
```
seems valid here

lib/voyager_web/live/settings_live/pid_format_settings.ex
```diff
@@ -0,0 +1,138 @@
...
+                  <span class="text-base-content font-medium">{text}</span>
+                  <kbd class="font-mono">{pid_string}</kbd>
+                </div>
+                <p class="text-base-content/60 text-xs">
+                  {raw(description)}
+                </p>
+              </div>
+            </li>
+          </ul>
+        </div>
```
piotrsac: - perhaps we could use a little bit more descriptive name than just data
- `raw/1` seems like a code smell to me

Perhaps we can just drop `@data` and do two `<li>`s rather than adding logic, i would just unroll it

lib/voyager_web/live/settings_live.ex
```diff
@@ -32,6 +37,7 @@ defmodule VoyagerWeb.SettingsLive do
       </div>
 
       <AppearanceSettings.appearance_settings />
+      <.live_component module={PidFormatSettings} id="pid-format-settings" pid_format={@pid_format} />
```
piotrsac: does that module actually have strong, specific need for it to be a `<.live_component>`?
maybe it is possible to do it like AppearanceSettings?

lib/voyager_web/formatters.ex
```diff
@@ -124,4 +124,43 @@ defmodule VoyagerWeb.Formatters do
...
+      ...>   pid |> :erlang.pid_to_list() |> List.to_string()
+      true
+      iex> VoyagerWeb.Formatters.format_pid("<123.23.423>", :local)
+      "<0.23.423>"
+  """
+  @spec format_pid(
+          pid() | String.t(),
+          :distribution | :local
+        ) :: String.t()
+  def format_pid(pid, format \\ default_pid_format())
```
piotrsac: `Formatters.default_pid_format/0` is a settings fallback, not presentation. `format_pid/2` belonging here is fine; this named default is not — SettingsLive and SupervisionTreeLive only reach for it as `Settings.get/2`’s default.

I’d drop the function and inline `:distribution` at the `Settings.get` call sites (same as MCP port, telemetry, suffix). Keep `format_pid(pid, format \\ :distribution)` as the formatter’s own omitted-arg default.

assets/js/hooks/supervision_tree/graph.js
```diff
@@ -218,6 +228,31 @@ export const graphMethods = {
...
+        node.data('displayLabel', composeLabel(node.data(), this.pidFormat));
+      });
+    });
+
+    if (this.nodeId) this.fillTooltip();
+    if (typeof this.repositionOverlays === 'function') {
+      this.repositionOverlays();
+    }
+  },
+
```
piotrsac: could we test that if it actually relabels the pids?

piotrsac:
Supervision-tree error banner still `{inspect(err)}`, so walk errors like `{:which_children, pid, …}` stay `#PID<…>` and ignore the toggle.

Not sure if that's intended or not

piotrsac (commented):
Changes are looking good and seem to work!

assets/js/hooks/supervision_tree/elements.js
GuzekAlan: Yes it is the same logic but i don't see any good way to share a code between 2 languages. Passing formatted string was a viable option but the select event would not be properly handled.

test/voyager_web/live/supervision_tree_live/details_panel_test.exs
GuzekAlan: i actually want to check that in all html there is a specific version and in all html there is no other. The docs means that the checking on raw element might not check the whole but I believe that if we want to refute that there isn't any code then it seems ok to me. Can you elaborate on the quote form agents.md and why to use it here if you still think it is valid ?

lib/voyager_web/live/settings_live.ex
GuzekAlan: It is a form with `handle_event`. `AppearanceSettings` has JS used for logic after selecting option in form

lib/voyager_web/formatters.ex
GuzekAlan: wouldn't it be a copied logic with multiple `default` values across whole project which needs to be changed if we decide to change the default setting ?

GuzekAlan: If it is to be consistent with the rest of the codebase I understand and don't object to doing it like you've suggested

GuzekAlan:
Regarding

>Supervision-tree error banner still {inspect(err)}, so walk errors like {:which_children, pid, …} stay #PID<…> and ignore the toggle.
> Not sure if that's intended or not

It might be good idea to leave errors as they are (easier to search in terminal) so I would leave them unformatted as they are now

hhubert6 (commented):
Have you considered using `:persistent_term` and just check in it with each format call? Also I don't think reacting to change in settings via pubsub everywhere is that necessary but adds logic which has to be added in many places later which is a little pain.

test/voyager_web/live/supervision_tree_live/details_panel_test.exs
piotrsac: wdyt about something like that?
```suggestion
      document = LazyHTML.from_fragment(html)
      name = document |> LazyHTML.query_by_id("details-panel-name") |> LazyHTML.text()

      assert name =~ "<0.45.0>"
      refute LazyHTML.text(document) =~ "<123.45.0>"
```

hhubert6:
@claude review opus

PR #191, opened 2026-08-25
Bug: Call the pr-review skill explicitly in the review workflow

dazakdev:
The review on #184 ignored `.claude/skills/pr-review/SKILL.md` completely — no severity labels, no 🦀 summary block, inline comments several paragraphs long, and a line explaining that no Elixir toolchain was available (the skill forbids exactly that).

## Cause

`track_progress: true` does not pass `prompt:` as the user prompt. The action wraps it in `<custom_instructions>` inside its own template:

```
<custom_instructions>
/pr-review REPO: software-mansion/voyager PR NUMBER: 184
...
</custom_instructions>
===== USER REQUEST =====
review opus
```

`/pr-review` sitting in the middle of a text block is not a slash command, so nothing expanded it. Whether the skill got loaded was the model's own call, prompted only by seeing the string and having `Skill` on the allowlist.

That held for #182, #159, #176 and #181 — all four followed the skill exactly, with byte-identical workflow config. It broke on #184, where the output matches the action's built-in template point for point: `###` headers, the todo-summary shape, and item (f) of the template ("if you are unable to run a linter or test suite, explain this in your comment").

## Fix

Call the Skill tool by name, and state that the skill wins on output format — otherwise the template still asks for h3 headers, a checklist and a what-I-could-not-do note after the skill loads.

## Not done

`show_full_output: true` would make the tool calls visible in the job log and settle this kind of question directly, instead of inferring from the posted comment. Left off — it prints the full agent output to the CI log. Worth adding if this recurs.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -47,6 +47,7 @@ jobs:
           github_token: ${{ github.token }}
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           track_progress: true
+          show_full_output: true
           prompt: |
             Call the Skill tool with `pr-review` NOW, before anything else, and follow it
             exactly. REPO: ${{ github.repository }} PR NUMBER: ${{ github.event.issue.number || github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


dazakdev:
@claude review

GuzekAlan (approved):
🐬

PR #195, opened 2026-08-27
Add ETS list API

piotrsac:
Closes [VOY-226](https://linear.app/softwaremansion/issue/VOY-226/ets-table-list-api). Cheap ETS **table metadata** over OTP MFA — no `:voyager_agent`, no inject.

## How

**Remote calls stay on `Voyager.Erpc`.** `format_error/2` + `safe_call/5` map timeout / noconnection / remote exceptions into `{:ok, _} | {:error, _}` so list/info (and later callers) do not each catch `:erpc`. Every call takes a timeout (default 5s).

**`Remote.list/2` is two bounded MFAs after `:ets.all/0`.** Empty `all` skips the rest. Otherwise: target `:erlang.system_info(:wordsize)`, then one `:lists.map(&:ets.info/1, ids)` so `info` runs on the target. Rows that die mid-fetch (`:undefined`) or are malformed are dropped; a length mismatch is `{:error, :invalid_response}`. Handle is the `all/0` id (name atom or live `reference()`), not the tid inside the info list. `memory` is **bytes** (`words * target wordsize`). Private tables stay in the list with `protection: :private`. `info/3` is the same shape for one table (`:undefined` → `:not_found`).

**`TableId` is the `?table=` / MCP string story.** Display is `inspect/1`. Resolve prefers the last `all`. Unnamed tables are never rebuilt with `list_to_ref` on the host. Typed names go through `:erlang.list_to_existing_atom/1` **on the target** (leading `:` stripped); never `String.to_atom/1` on Voyager.

**Tests** use `Voyager.ErpcMock` for MFA, timeouts, and error mapping. A small live suite swaps in `Erpc.Impl` and lists real local tables (including private / unnamed).


piotrsac:
@claude review opus

piotrsac:
@claude review opus

lib/voyager/services/ets/table_id.ex
```diff
@@ -0,0 +1,127 @@
...
+  Interns `name` as an existing atom **on `node`**.
+
+  A leading Elixir colon (`":foo"`) is stripped so inspect-forms of simple
+  atoms can be typed without a prior `all`. Alias inspect-forms (`"MyApp.Cache"`)
+  are interned as `:"Elixir.MyApp.Cache"`; a string that already starts with
+  `"Elixir."` is left as-is. Returns `{:error, :not_found}` when the atom is
+  not interned on the target.
+  """
+  @spec existing_atom(node(), String.t(), timeout()) :: {:ok, atom()} | {:error, term()}
+  def existing_atom(node, name, timeout) when is_binary(name) do
```
GuzekAlan, line 83: 🟡 should-fix: `existing_atom/3` puts no length bound on `name`, but an atom can never exceed 255 characters — a multi-megabyte `?table=`/MCP string gets charlist-ified and shipped over distribution to the observed node only to fail `badarg` there. Reject names longer than 255 characters host-side as `{:error, :invalid_name}`, next to the existing blank-name checks.

---

lib/voyager/services/ets/remote.ex
```diff
@@ -0,0 +1,161 @@
...
+
+  `:ets.info/1` includes **private** tables; they are flagged via
+  `protection: :private` and are not omitted. `memory` is converted to
+  **bytes** using the target's `:erlang.system_info(:wordsize)`.
+  """
+
+  alias Voyager.Erpc
+  alias Voyager.Services.Ets.TableId
+
+  @default_timeout 5_000
```
GuzekAlan: 🟡 should-fix: `@default_timeout 5_000` is a second copy of the erpc default timeout — `lib/voyager/services/process_info.ex:12` has the same `@timeout 5_000`. Now that `Voyager.Erpc.safe_call/5` is the shared seam, give the default one home there (e.g. a `safe_call/4` defaulting to 5s) so the two values cannot drift.

---

GuzekAlan (commented):
## 🦀 Review

- 🟡 `lib/voyager/services/ets/table_id.ex:86` — `existing_atom/3` has no length bound; atoms cap at 255 characters, so a longer `?table=`/MCP string is a guaranteed remote `badarg`, yet the whole string is charlist-ified and sent over distribution to the observed node first. Reject `> 255` characters host-side as `:invalid_name`.
- 🟡 `lib/voyager/services/ets/remote.ex:17` — `@default_timeout 5_000` duplicates `@timeout 5_000` in `lib/voyager/services/process_info.ex:12`; with `safe_call/5` now the shared seam, the erpc default timeout belongs in one place in `Voyager.Erpc`.

<verdict: 0 blocking, 2 should-fix, 0 nit>

---

lib/voyager/erpc.ex
```diff
@@ -30,6 +30,35 @@ defmodule Voyager.Erpc do
...
+  def format_error(:throw, value), do: {:error, {:remote_throw, value}}
+
+  @doc """
+  Like `call/5`, but returns `{:ok, result}` or `{:error, reason}` instead of
+  raising.
+  """
+  @spec safe_call(node(), module(), atom(), [term()], timeout_time() | call_options()) ::
+          {:ok, term()} | {:error, term()}
+  def safe_call(node, mod, fun, args, timeout_or_options) do
+    {:ok, call(node, mod, fun, args, timeout_or_options)}
```
GuzekAlan: could you also add `safe_call/4` ?

lib/voyager/services/ets/remote.ex
```diff
@@ -0,0 +1,161 @@
...
+           Erpc.safe_call(node, :erlang, :system_info, [:wordsize], timeout) do
+      case build(table, info, word_size) do
+        {:ok, table_info} -> {:ok, table_info}
+        :error -> {:error, :invalid_response}
+      end
+    else
+      {:ok, :undefined} -> {:error, :not_found}
+      {:ok, _} -> {:error, :invalid_response}
+      {:error, _} = err -> err
+    end
```
GuzekAlan, line 69: why not use `fetch_infos(node, [table], timeout)` ?

lib/voyager/services/ets/table_id.ex
```diff
@@ -0,0 +1,136 @@
...
+
+  @max_atom_chars 255
+
+  @type t :: atom() | reference()
+
+  @doc """
+  Stable display form of a table handle, suitable for `?table=` and MCP.
+  """
+  @spec display(t()) :: String.t()
+  def display(id) when is_atom(id) or is_reference(id), do: inspect(id)
```
GuzekAlan: Nit: you can add `defguard` for `is_table_id`

lib/voyager/services/ets/table_id.ex
```diff
@@ -0,0 +1,136 @@
+defmodule Voyager.Services.Ets.TableId do
+  @moduledoc """
+  Identifies ETS tables across list fetches, `?table=` params, and MCP strings.
+
+  A table handle is a **name atom** or a live **`reference()`** (unnamed tables).
+  Display and matching use `inspect/1`. Unnamed tables cannot be rebuilt on the
+  Voyager host — never `:erlang.list_to_ref/1` — so a reference inspect-string
+  is resolved only against the last `:ets.all/0`. Typed names are interned on
+  the *target* with `:erlang.list_to_existing_atom/1`, never `String.to_atom/1`
+  on the host.
```
GuzekAlan: Nit: Some comments look like copy paste from linear issue. Can you leave only necessary info ?

lib/voyager/services/ets/remote.ex
piotrsac: `fetch_infos/3` is the list path: best-effort, drop vanished/malformed rows, return `{:ok, list}`. `info/3` is a lookup — `:undefined` is `:not_found`, a bad keyword list is `:invalid_response`, and we skip wordsize when the table is already gone.

Reusing it as `fetch_infos(node, [table], timeout)` would map both of those to `{:ok, []}` (and always fetch wordsize first, via `:lists.map` instead of `:ets.info/1`). Sharing `build/3` is the overlap that actually matches.

GuzekAlan (commented):
Have you considered `list_to_reference` ?

piotrsac:
There is no `list_to_reference/1` — the BIF is `erlang:list_to_ref/1`, and OTP marks it debug-only.

We looked at it as the unnamed-table analog of `list_to_existing_atom/1`. It is not a better fallback:
- `inspect/1` is `#Reference<…>`; `list_to_ref` only accepts `#Ref<…>` and raises otherwise.
- A `#Ref<0.…>` from the target’s shell reconstructed on Voyager is a local Voyager ref. ETS tables are node-local, so that handle will not find the table.
- `list_to_ref` on a node with fewer schedulers than the origin can `badarg`.

Named tables can be interned on the target because atoms exist there. Unnamed tids are the live `reference()` values from `:ets.all/0`; resolve matches that list instead of forging a new one.

GuzekAlan (approved):
🐙

PR #200, opened 2026-08-28
Feature: Mark locally built apps with a "Dev Build" banner

GuzekAlan:
Closes [VOY-164](https://linear.app/softwaremansion/issue/VOY-164/add-dev-banner).

A locally built desktop app looked exactly like a released one. This adds a strong-colored `Dev Build` banner above the topbar on local builds only.

## How it is switched on

`rel/app/tauri.sh` already sources `rel/app/.env` (git-ignored) for `dev` and `app`; it now does so for every command, so a local build is configured the same way it is run. CI checks out no `.env`, which is what keeps published builds unmarked — no config file or code change is needed to toggle it:

```sh
# rel/app/.env, copied from rel/app/.env.sample
VOYAGER_DEV_BUILD=true
```

The value is read in `config/config.exs` rather than `runtime.exs` on purpose: the bundled desktop app is launched through the OS (`open`), so it does not inherit the shell environment it was built in. Reading it at build time bakes the flag into the release, while `mix phx.server` still picks it up from the environment at boot. Only `1`/`true` enable it, so the blank entry in `.env.sample` reads as off.

## Where the banner lives

It is rendered from the root layout so it tops every page, including connect and settings, which have no app shell. The layouts below it move from `h-screen` to `h-full` so they size to the space left over instead of to the full viewport — verified there is no page overflow with the banner on (`scrollHeight == clientHeight`).

`text-neutral` is used instead of `text-warning-content`, which is tuned for the muted `warning-bg` alert surface and washes out on solid amber; `text-neutral` stays dark in both themes.

## Testing

- `mix precommit` — green on the merged tree (383 tests, plus the Rust suite).
- New `test/voyager_web/components/layouts_test.exs` covers the component in both states and asserts the rendered page carries / omits `#dev-build-banner` on both `/` and `/settings` — one per `live_session`, since each composes its own layout with the shared root.
- Rendered `/` and `/settings` in Chromium in both light and dark themes with the flag on: banner shows above the topbar, contrast holds in both themes, no page overflow.

> [!NOTE]
> The one red `Build and test` run on 4d42126 was an unrelated port race in `Voyager.MCPCase` — [diagnosis here](https://github.com/software-mansion/voyager/pull/200#issuecomment-5454019179). Green again since `main` was merged in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,19 @@ mix assets.deploy
 mix tauri.app
 ```
 
+### Marking local builds
+
+`rel/app/tauri.sh` sources `rel/app/.env` (git-ignored) before building or
+running the desktop app. Setting `VOYAGER_DEV_BUILD` there puts a `Dev Build`
+banner above the topbar, so a locally built app is never mistaken for a release:
+
+```sh
+# rel/app/.env
+VOYAGER_DEV_BUILD=true
+```
+
+CI has no `rel/app/.env`, so published builds never carry the banner.
+
 Before opening a pull request, run:
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -122,9 +122,10 @@ mix tauri.app
 
 ### Marking local builds
 
-`rel/app/tauri.sh` sources `rel/app/.env` (git-ignored) before building or
-running the desktop app. Setting `VOYAGER_DEV_BUILD` there puts a `Dev Build`
-banner above the topbar, so a locally built app is never mistaken for a release:
+`rel/app/tauri.sh` sources `rel/app/.env` (git-ignored, copied from
+[`rel/app/.env.sample`](rel/app/.env.sample)) before building or running the
+desktop app. Setting `VOYAGER_DEV_BUILD` there puts a `Dev Build` banner above
+the topbar, so a locally built app is never mistaken for a release:
 
 ```sh
 # rel/app/.env
--- a/config/config.exs
+++ b/config/config.exs
@@ -6,6 +6,13 @@ config :voyager,
 
 config :voyager, :terms_of_service_url, "https://swmansion.com/legal/voyager/terms-of-service/"
 
+# Marks locally built apps with a "Dev Build" banner. Read at build time rather
+# than in `runtime.exs`, because the bundled desktop app is launched without the
+# shell environment it was built in. `rel/app/tauri.sh` exports this from
+# `rel/app/.env`, a file that only exists on developer machines, so builds made
+# on CI are never marked.
+config :voyager, :dev_build?, System.get_env("VOYAGER_DEV_BUILD") in ~w(1 true)
+
 config :voyager, VoyagerWeb.Endpoint,
   url: [host: "localhost"],
   adapter: Bandit.PhoenixAdapter,
--- a/lib/voyager_web/components/layouts.ex
+++ b/lib/voyager_web/components/layouts.ex
@@ -69,13 +69,36 @@ defmodule VoyagerWeb.Layouts do
   def settings(assigns) do
     ~H"""
     <.flash_group flash={@flash} />
-    <div class="bg-base-200 flex h-screen flex-col overflow-hidden">
+    <div class="bg-base-200 flex h-full flex-col overflow-hidden">
       <VoyagerWeb.Components.Shell.settings_topbar return_to={assigns[:return_to]} />
       <main class="flex-1 overflow-y-auto">
         {@inner_content}
       </main>
     </div>
     <.onboarding_modal show={assigns[:show_onboarding?]} />
+    """
+  end
+
+  @doc """
+  Renders the "Dev Build" banner that tops every page of a locally built app.
+
+  Renders nothing unless `Voyager.dev_build?/0` is true, so released builds
+  never show it.
+  """
+  def dev_banner(assigns) do
+    assigns = assign(assigns, :dev_build?, Voyager.dev_build?())
+
+    ~H"""
+    <%!-- `text-neutral` rather than `text-warning-content`: the latter is tuned
+          for the muted `warning-bg` alert surface and washes out on solid amber. --%>
+    <div
+      :if={@dev_build?}
+      id="dev-build-banner"
+      role="status"
+      class="bg-warning text-neutral tracking-label flex flex-none items-center justify-center gap-1.5 px-4 py-1 text-xs font-semibold uppercase"
+    >
+      <.icon name="icon-circle-alert" class="size-3.5" /> Dev Build
+    </div>
     """
   end
 
--- a/test/voyager_web/components/layouts_test.exs
+++ b/test/voyager_web/components/layouts_test.exs
@@ -1,0 +1,52 @@
+defmodule VoyagerWeb.LayoutsTest do
+  use VoyagerWeb.ConnCase, async: false
+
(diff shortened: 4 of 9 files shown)


GuzekAlan:
@claude review opus

---

test/voyager_web/components/layouts_test.exs
```diff
@@ -0,0 +1,52 @@
...
+    test "leaves the page unmarked on a released build", %{conn: conn} do
+      Application.put_env(:voyager, :dev_build?, false)
+
+      assert conn |> render_page() |> banner_count() == 0
+    end
+  end
+
+  defp render_page(conn) do
+    conn |> get(~p"/") |> html_response(200) |> LazyHTML.from_document()
+  end
```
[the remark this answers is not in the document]
GuzekAlan: Good catch — the settings layout is one of the two this PR resized from `h-screen` to `h-full`, so it was the least covered part of the change. Fixed in 293ba1b: the root-layout assertions now run against both `/` (connect) and `/settings`, generated over a `@pages` list so adding a future `live_session` is a one-line change.

---

GuzekAlan:
Both reviews addressed.

**Claude — 🟢 `rel/app/.env.sample`:** correct, and I'd missed that the file exists. Fixed in b839cc0: `VOYAGER_DEV_BUILD=` now sits alongside the telemetry vars, left blank so the sample stays off by default (only `1`/`true` enable the banner). The README section now points at the sample rather than describing `.env` as something you write from scratch.

**Copilot — settings-page test coverage:** fixed in 293ba1b, replied on the thread.

`mix precommit` is green on the current head.

---

test/voyager_web/components/layouts_test.exs
```diff
@@ -0,0 +1,58 @@
...
+    on_exit(fn -> Application.put_env(:voyager, :dev_build?, previous) end)
+    :ok
+  end
+
+  describe "dev_banner/1" do
+    test "renders nothing on a released build" do
+      Application.put_env(:voyager, :dev_build?, false)
+
+      assert render_component(&Layouts.dev_banner/1) =~ ~r/\A\s*\z/
+    end
```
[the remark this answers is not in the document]
GuzekAlan: Taking the two halves separately — one implemented, one deliberately not. Leaving this unresolved so you can weigh in.

**Line 27 (`assert html =~ "Dev Build"`) — agreed, fixed in 4d42126.** You're right that matching the string anywhere in the rendered output could be satisfied by unrelated markup. It now reads the text off the selected node:

```elixir
banner = render_banner()

assert Enum.count(banner) == 1
assert banner |> LazyHTML.text() |> String.trim() == "Dev Build"
```

**Line 19 — keeping an emptiness assertion, not a selector one.** The proposed change (parse and assert `#dev-build-banner` is absent) is strictly weaker than what's there: it would still pass if the component emitted a stray wrapper, a comment node, or a differently-named element on a released build. What this test is actually pinning down is that `dev_banner/1` contributes *nothing at all* to a release — that's the property that matters, since the whole point is that a published build carries no trace of the banner. I did drop the regex for something plainer:

```elixir
assert render_component(&Layouts.dev_banner/1) |> String.trim() == ""
```

`CLAUDE.md:458` reads to me as a rule against asserting on HTML *structure or content* in raw strings — coupling a test to markup that can legitimately change. Asserting that no markup was produced isn't that; there is no structure to select on, and any selector-based phrasing of it would assert less. The positive case is where the rule bites, and that one is now selector-based.

`mix precommit` is green on the current head.

---

GuzekAlan:
### `Build and test` went red on 4d42126 — not this PR's, and worth a follow-up

**What failed:** `Voyager.MCP.TelemetryVerificationTest` — "subscribed MCP events fire on a real tools/call", with `(Req.TransportError) connection refused`. The real cause is a line above it in the log:

```
[error] Running Voyager.MCP.Router with Bandit 1.12.5 at http failed, port 56276 already in use
[warning] MCP endpoint failed to start on port 56276: :port_in_use
```

**Why it isn't this PR's:** the diff touches the root layout, `config/config.exs`, `rel/app/tauri.sh`, the README and one layout test — nothing in MCP, Bandit, or port handling. The commit that went red (`4d42126`) changed only assertions inside `layouts_test.exs`, and this same check passed on the two commits before it on this branch.

**Root cause, for whoever wants it:** `Voyager.MCPCase.unique_port/0` picks `54_000 + rem(System.unique_integer(), 4_000)`, i.e. **54000–57999**. On the Linux runner `net.ipv4.ip_local_port_range` is **32768–60999**, so that window sits entirely inside the ephemeral range and the kernel can hand the same port to an outbound socket before Bandit binds it. 56276 is squarely in the overlap. It's a genuine race rather than an infra blip, and it will keep biting at random across PRs.

I haven't fixed it here — it's in shared test-support code with no relation to a dev-build banner, so it belongs in its own change rather than widening this PR. The fix is small if someone wants it: move the range above `ip_local_port_range` (e.g. base 61000+), or bind port `0` and read back the assigned port instead of guessing one. Happy to open that separately.

**What I did instead:** merged `main` (which had moved to 8a7b800) into the branch and pushed `f1bf076`, which re-runs the check. Locally on the merged tree the full suite is green — 383 passed, the same count CI runs — including the MCP test.

---

dazakdev:
@claude review

hhubert6 (approved):
🐬

lib/voyager_web/components/layouts.ex
```diff
@@ -79,6 +79,21 @@ defmodule VoyagerWeb.Layouts do
...
 
+  def dev_banner(assigns) do
+    assigns = assign(assigns, :dev_build?, Application.get_env(:voyager, :dev_build?, false))
+
+    ~H"""
+    <div
+      :if={@dev_build?}
+      id="dev-build-banner"
+      role="status"
+      class="bg-warning-bg text-warning-content border-warning/30 tracking-label font-mono flex flex-none items-center justify-center gap-1.5 border-b px-4 py-1 text-xs uppercase"
```
piotrsac:
```suggestion
      class="bg-warning-bg text-warning-content border-warning/30 tracking-label font-mono flex flex-none items-center justify-center gap-2 border-b px-4 py-2 text-xs uppercase"
```

README.md
```diff
@@ -124,6 +124,10 @@ mix assets.deploy
 mix tauri.app
 ```
 
+Development runs always show a `Dev Build` banner. Production builds (`mix tauri.app`,
+`mix tauri.build`) are built with `MIX_ENV=prod`, so set `VOYAGER_DEV_BUILD=true` in `rel/app/.env`
+(see [`.env.sample`](rel/app/.env.sample)) to mark them too.
```
piotrsac: > so set `VOYAGER_DEV_BUILD=true` in `rel/app/.env` [...] to mark them too.

i dont really see the point and logic in that part of the sentence

config/config.exs
```diff
@@ -6,6 +6,9 @@ config :voyager,
 
 config :voyager, :terms_of_service_url, "https://swmansion.com/legal/voyager/terms-of-service/"
 
+# Not in `runtime.exs`: the bundled desktop app starts without the shell environment it was built in.
```
piotrsac:
```suggestion
```

piotrsac (approved):
🦭

PR #205, opened 2026-08-31
Fix: Generate random distribution cookie per launch

dazakdev:
Fixes [VOY-219](https://linear.app/softwaremansion/issue/VOY-219/distribution-cookie-is-shipped-in-the-release-and-identical-across-all).

## Problem

Every release build ships the same `releases/COOKIE`: identical across all installs of a version, and passed via `-setcookie` in argv (`ps`-visible).
## Fix (two layers, both needed)

1. **`rel/app/env.sh.eex`** — export a fresh random `RELEASE_COOKIE` on every launch. `env.sh` is sourced before the `releases/COOKIE` fallback in the generated bin script, so the shipped file is never used.
2. **`Voyager.Services.Distribution`** — rotate the cookie in-VM (`Node.set_cookie/1` with 32 random bytes) immediately after `net_kernel` starts, on both the `:ok` and `{:already_started, _}` paths. Kills the argv/`ps` leak: the boot cookie is dead the moment the node is connectable.

Layer 1 alone leaves the per-launch cookie `ps`-readable for the whole session; layer 2 alone leaves a race window where the publicly-known shipped cookie is accepted.

Outbound connections are unaffected — both connectors set per-peer cookies (`:erlang.set_cookie(node, cookie)`) before `Node.connect`. Side effect (intended): `bin/voyager remote`/`rpc` can no longer attach to a running app; nothing uses them.

## Verification

- Built prod release, launched twice with the same shipped `COOKIE` file: two different runtime cookies, neither matching the shipped one.
- New test asserts the cookie rotates across distribution restarts and is never `:nocookie`.
- `mix precommit` green.

Skipped as not shipped: `env.bat.eex` parity (no Windows build in CI), removing the now-inert bundled `COOKIE` file.


GuzekAlan:
@claude review opus

srzeszut:
I would delete node cookie as well in connectors.
```
      case Node.connect(remote_node) do
        true ->
          :erlang.set_cookie(remote_node, :nocookie)
          {:ok, remote_node, conn_ref, local_port}

```
like this (the same in distribution connector)
Now, we can peek cookie even after disconect.

srzeszut:
@claude review

dazakdev:
@claude review

srzeszut:
@claude review

hhubert6 (approved):
Nice 🐬

PR #208, opened 2026-08-31
Feature: Process list page

GuzekAlan:


lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,439 @@
...
+  def handle_event("restore_settings", params, socket) do
+    socket
+    |> assign(:page_size, page_size(parse_integer(params["page_size"])))
+    |> apply_controls(params)
+    |> fetch()
+    |> noreply()
+  end
+
+  def handle_event("sort", %{"key" => key}, socket) do
+    key = String.to_existing_atom(key)
```
GuzekAlan: 🔴 blocking: `String.to_existing_atom/1` on a client-supplied `key` — a `sort` event with any string that is not an interned atom raises and kills the LiveView (reproduced with `render_hook(view, "sort", %{"key" => "nope_123"})`). Compare as strings against `Processes.sortable_attrs()` and ignore anything else.

---

lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,439 @@
...
+    page = parse_integer(page) || socket.assigns.page
+
+    socket
+    |> assign(:page, clamp_page(socket, page, socket.assigns.page_size))
+    |> noreply()
+  end
+
+  def handle_event("set_interval", %{"interval" => value}, socket) do
+    socket
+    |> assign(:refresh_interval, parse_interval(value))
```
GuzekAlan: 🔴 blocking: `set_interval` accepts any integer. `"-5"` crashes the LiveView in `Process.send_after/3` (reproduced), and `"0"`/`"1"` turn `:auto_refresh` into a tight loop that spins the LiveView and the rate limiter. Validate the value against `Processes.refresh_interval_options()` and ignore anything else.

---

lib/voyager_web/form_schemas/process_list_controls.ex
```diff
@@ -0,0 +1,124 @@
...
+  @spec default() :: t()
+  def default do
+    %__MODULE__{columns: Enum.map(Processes.default_attrs(), &to_string/1)}
+  end
+
+  @spec changeset(t(), map()) :: Ecto.Changeset.t()
+  def changeset(controls \\ default(), attrs \\ %{}) do
+    controls
+    |> cast(attrs, [:search, :limit, :timeout, :columns])
+    |> update_change(:search, &String.trim/1)
```
GuzekAlan, line 60: 🔴 blocking: `update_change/3` still runs when the change is `nil`, so `String.trim(nil)` raises. `restore_settings` is pushed from the browser with whatever localStorage holds; `{"search": null}` kills the LiveView (reproduced). Trim with a nil-safe function (`&String.trim(&1 || "")`) or reject non-binaries in `restore_settings` before applying.

---

lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,439 @@
...
+    |> assign(:sort_by, sort_by)
+    |> assign(:direction, direction)
+    |> assign(:page, 1)
+    |> assign(:page_size, @default_page_size)
+    |> assign(:page_sizes, @page_sizes)
+    |> assign(:refresh_interval, nil)
+    |> assign(:refresh_timer, nil)
+    |> assign(:last_updated, nil)
+    |> assign(:round_trip_ms, nil)
+    |> then(&if(connected?(&1), do: fetch(&1), else: &1))
```
GuzekAlan: 🟡 should-fix: with stored settings every visit runs two full scans. Mount fetches with the defaults, then the hook's `restore_settings` fetches again; the second `start_async` discards the first result, but the erpc-spawned scan still runs to completion on the node. Traced on a 100k-process node: 2 `proc_top` calls per navigation, reload and reconnect. Let the hook always push `restore_settings` (empty when nothing is stored) and fetch only from there, or skip the mount fetch when the hook is present.

---

lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,439 @@
...
+  # with the remembered options rather than the defaults.
+  def handle_event("restore_settings", params, socket) do
+    socket
+    |> assign(:page_size, page_size(parse_integer(params["page_size"])))
+    |> apply_controls(params)
+    |> fetch()
+    |> noreply()
+  end
+
+  def handle_event("sort", %{"key" => key}, socket) do
```
GuzekAlan, line 154: 🟡 should-fix: `sort` (unlike `:auto_refresh`) has no in-flight guard, and a replaced `start_async` does not stop the scan already spawned on the remote. Measured: 24 header clicks in 5 s produced 24 overlapping full scans on a 100k-process node, per-scan time went from 435 ms to 2 s and the node's run queue from 1 to 9; the rate limiter's 10-token burst plus 5/s refill lets that through. Ignore `sort`/`refresh_now`/`restore_settings` while `page_result.loading`, and disable the header buttons while loading.

---

lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,439 @@
...
+  # Rate limiting and timeouts are transient, and the rows already on screen are
+  # still the last good answer, so both flash rather than replacing the table.
+  def handle_async(:page_result, {:ok, {:rate_limited, _retry_after_ms}}, socket) do
+    socket
+    |> clear_loading()
+    |> put_flash(:error, "Too many requests.")
+    |> noreply()
+  end
+
+  def handle_async(:page_result, {:ok, {:error, :timeout}}, socket) do
```
GuzekAlan: 🟡 should-fix: when the *first* fetch times out there are no rows to keep, so this clause leaves the page on "Scanning processes…" with `waiting for first fetch…` and only a toast. Reproduced on a loaded 200k-process node (scan 10 s, default timeout 5 s): the table never leaves the loading text and nothing tells the user to raise the timeout. When `page_result.ok?` is false, set `AsyncResult.failed` (with a "raise the timeout" hint) instead of clearing loading.

---

lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,439 @@
...
+    socket
+    |> assign(:page_result, AsyncResult.loading(socket.assigns.page_result))
+    |> start_async(:page_result, fn ->
+      run_fetch(priority, session.node, controls, {sort_by, direction})
+    end)
+  end
+
+  defp run_fetch(priority, node, controls, sort) do
+    started = System.monotonic_time(:millisecond)
+    result = RateLimiter.run(priority, fn -> Processes.page(node, controls, sort) end)
```
GuzekAlan: 🟡 should-fix: the timeout only bounds how long Voyager waits; `:erpc.call/5` abandons the remote worker, which keeps scanning. Traced: 5 s timeout, `proc_top` returned after 9.8 s on the node, and each retry (or the next auto-refresh tick, which is not blocked by a timed-out fetch) costs the node another full scan. Either kill the remote worker on timeout (spawn the scan via `spawn_request` with a monitor and `exit(Pid, kill)` when the wait ends) or make the "Timeout (ms)" help text say it does not bound work on the node.

---

lib/voyager_web/components/core_components.ex
```diff
@@ -619,7 +716,6 @@ defmodule VoyagerWeb.CoreComponents do
       <div
         id={"#{@id}-tip"}
         role="tooltip"
```
GuzekAlan, line 719: 🟡 should-fix: dropping `phx-update="ignore"` from the tip means any re-render of its content wipes the state the `Tooltip` hook put on it (`is-open`, `is-pinned`, inline `top`/`left`). Reproduced: pin a reductions cell tooltip, click refresh, the tooltip is gone; with 1 s auto-refresh a hovered tooltip vanishes every tick. This changes every tooltip in the app, not only the table. Keep `ignore` on the tip and have the hook refresh the text it needs, or re-apply the open/pinned state in the hook's `updated()`.

---

lib/voyager_web/components/process_components.ex
```diff
@@ -0,0 +1,352 @@
...
+  attr :id, :string, required: true
+  attr :value, :string, required: true
+  attr :muted, :boolean, default: false
+  attr :class, :any, default: nil
+
+  def value_cell(assigns) do
+    assigns = assign(assigns, :empty?, assigns.value == @placeholder)
+
+    ~H"""
+    <.tooltip id={"#{@id}-tip"} interactive class="min-w-0 max-w-full" tip_class="font-mono">
```
GuzekAlan: 🟡 should-fix: a `Tooltip` hook plus a portaled copy of the value for every cell. At 100 rows that is 714 hooks, 714 portal nodes and ~9k DOM nodes, all re-patched on each refresh (a 300 ms scan renders in ~900 ms end to end). The ETS page will multiply this. A `title` attribute for the full value plus one copy button per row, or a single delegated tooltip for the table, gives the same affordance without a hook per cell.

---

lib/voyager_web/components/data_table_components.ex
```diff
@@ -0,0 +1,265 @@
...
+  the card rather than spilling past its border.
+  """
+  @spec page_min_width_class() :: String.t()
+  def page_min_width_class, do: "min-w-6xl"
+
+  @doc """
+  Sortable table.
+
+  `rows` is a list of `{dom_id, row}` tuples so the caller can drive it from a
+  LiveView stream or a plain list. Each column renders through the `:cell` slot,
```
GuzekAlan: 🟡 should-fix: the doc says `rows` can come from a LiveView stream, but the `<tbody>` has no `phx-update="stream"` and the empty state is `:if={@rows == []}`; a stream is not enumerable, so neither works. The ETS page will need streams (`CLAUDE.md` requires them for collections), so either add a stream mode (`phx-update="stream"` on the tbody, `hidden only:block` empty row) or drop the claim.

---

lib/voyager_web/components/data_table_components.ex
```diff
@@ -0,0 +1,265 @@
...
+
+    ~H"""
+    <th
+      data-column={@column.key}
+      class={[align_class(@column), width_class(@column)]}
+      aria-sort={aria_sort(@active?, @direction)}
+    >
+      <button
+        type="button"
+        phx-click="sort"
```
GuzekAlan, line 103: 🟡 should-fix: the event names `sort`, `paginate` and `set_page_size` are now hard-coded, so two tables on one page (an ETS table list next to its records, for example) cannot be told apart, and the pager and table cannot be reused inside a LiveComponent without a `phx-target`. Keep the `sort_event`/`event` attrs (or accept a `target`) so the caller decides where events go.

---

lib/voyager/queries/processes.ex
```diff
@@ -0,0 +1,162 @@
...
+  never touch `:erlang.list_to_pid/1` themselves.
+
+  Ranking, searching and truncation all run remotely, so `page/3` returns at
+  most `limit` entries out of `scanned` walked processes. Paging is the
+  caller's: every fetched row comes back.
+  """
+
+  alias Voyager.Services.ProcessInfo
+  alias Voyager.Services.ProcessList
+  alias VoyagerWeb.FormSchemas.ProcessListControls
```
GuzekAlan: 🟡 should-fix: `Voyager.Queries.Processes` pattern-matches on `VoyagerWeb.FormSchemas.ProcessListControls`, and that schema calls back into `Voyager.Queries.Processes`, so `lib/voyager` now depends on `lib/voyager_web` in a cycle. The query module also carries UI-only option labels (`@refresh_intervals`, a copy of `NodeInfoLive.@interval_options`). Take plain arguments in `page/…` (attrs, limit, timeout, search, sort) and move the interval options next to `interval_select` so both pages share one list.

---

lib/voyager_web/components/process_components.ex
```diff
@@ -0,0 +1,352 @@
...
+          long error cannot stretch the grid column and shift the controls. --%>
+    <p class="font-mono text-error relative h-4 w-24 text-xs">
+      <span class="absolute left-0 whitespace-nowrap">
+        {@field.errors |> Enum.map_join(", ", &translate_error/1)}
+      </span>
+    </p>
+    """
+  end
+
+  defp translate_error({msg, opts}) do
```
GuzekAlan: 🟡 should-fix: `translate_error/1` and `field_error/1` duplicate what `core_components.ex` already has, because the form renders raw `<select>`/`<input>` instead of the `<.input>` component `CLAUDE.md` requires. `<.input type="select">` and `<.input type="number">` render the error for you and drop both helpers.

---

lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,439 @@
...
+  end
+
+  def handle_async(:page_result, {:ok, {:error, reason}}, socket) do
+    socket
+    |> assign(:page_result, AsyncResult.failed(socket.assigns.page_result, reason))
+    |> assign(:dirty?, false)
+    |> noreply()
+  end
+
+  def handle_async(:page_result, {:exit, reason}, socket) do
```
GuzekAlan: 🟢 nit: this clause is the only failure path that does not reset `dirty?`, so a crashed fetch started by `:refetch` leaves the table dimmed and `pointer-events-none` until a later fetch succeeds. Clear it as the other clauses do.

---

lib/voyager/node_session.ex
```diff
@@ -34,6 +34,22 @@ defmodule Voyager.NodeSession do
...
   end
 
+  @doc """
+  Restores the cached connector name, or clears it when given `nil`.
+
+  The cache is a `:persistent_term` and so outlives any process state, which
+  makes it leak across tests that connect; this lets them put it back.
+  """
+  @spec restore_cached_connector_name(atom() | nil) :: :ok
+  def restore_cached_connector_name(nil) do
```
GuzekAlan: 🟢 nit: a public function on `Voyager.NodeSession` whose only caller is a test. If it has to live here, `@doc false` keeps it out of the module's API; otherwise expose the cache key and keep the reset in `Voyager.Fakes`.

---

lib/voyager_web/form_schemas/process_list_controls.ex
```diff
@@ -0,0 +1,124 @@
...
+    |> update_change(:columns, &filter_known/1)
+    |> validate_inclusion(:limit, @limits, message: "must be one of #{Enum.join(@limits, ", ")}")
+    |> validate_number(:timeout,
```
GuzekAlan, line 64: 🟢 nit: clearing the timeout box submits `""`, which `cast` treats as no change, so the field stays empty on screen while the previous timeout is silently used and no error shows. `validate_required([:timeout, :limit])` makes the empty state visible.

---

GuzekAlan (commented):
## 🦀 Review

- 🔴 `lib/voyager_web/live/processes_live.ex:170` — `String.to_existing_atom/1` on the client-supplied sort `key`; any non-interned string kills the LiveView. Compare as strings against `Processes.sortable_attrs()`.
- 🔴 `lib/voyager_web/live/processes_live.ex:204` — `set_interval` takes any integer: `"-5"` crashes in `Process.send_after/3`, `"0"` spins `:auto_refresh` in a tight loop. Validate against `refresh_interval_options()`.
- 🔴 `lib/voyager_web/form_schemas/process_list_controls.ex:60` — `update_change(:search, &String.trim/1)` raises on `nil`; a `restore_settings` push with `"search": null` kills the LiveView. Trim nil-safely or reject non-binaries.
- 🟡 `lib/voyager_web/live/processes_live.ex:56` — mount fetch plus the hook's `restore_settings` fetch = two full remote scans on every visit, reload and reconnect when settings are stored; the first result is discarded but the scan still runs on the node.
- 🟡 `lib/voyager_web/live/processes_live.ex:169` — `sort`/`refresh`/`restore` have no in-flight guard and replaced async tasks do not stop the remote scan: 24 header clicks in 5 s → 24 overlapping scans on a 100k-process node, per-scan latency 435 ms → 2 s, run queue 1 → 9. Ignore while `page_result.loading`.
- 🟡 `lib/voyager_web/live/processes_live.ex:266` — a timed-out *first* fetch leaves the page on "Scanning processes…" with only a toast and no error state (200k processes under load: 10 s scan vs 5 s default timeout).
- 🟡 `lib/voyager_web/live/processes_live.ex:319` — the timeout bounds the wait, not the node: the erpc worker keeps scanning after Voyager gives up (5 s timeout, `proc_top` finished at 9.8 s), and each retry costs the node a full scan. Kill the remote worker on timeout or document the limit.
- 🟡 `lib/voyager_web/components/core_components.ex:718` — removing `phx-update="ignore"` from the tip resets the hook's open/pinned state and position on every re-render; a pinned cell tooltip disappears on refresh, and this affects every tooltip in the app.
- 🟡 `lib/voyager_web/components/process_components.ex:257` — a `Tooltip` hook and a portaled copy per cell: 714 hooks / ~9k DOM nodes at 100 rows, re-patched on every refresh. Use `title` plus one copy button per row, or a delegated tooltip.
- 🟡 `lib/voyager_web/components/data_table_components.ex:43` — claims stream support but the `<tbody>` has no `phx-update="stream"` and the empty row is `@rows == []`; the ETS page will need streams.
- 🟡 `lib/voyager_web/components/data_table_components.ex:114` — `sort`/`paginate`/`set_page_size` are hard-coded, so two tables on one page or a table inside a LiveComponent cannot route events. Keep the event/target attrs.
- 🟡 `lib/voyager/queries/processes.ex:18` — `Voyager.Queries.Processes` ↔ `VoyagerWeb.FormSchemas.ProcessListControls` is a `lib/voyager` ↔ `lib/voyager_web` dependency cycle, and the query module carries UI option labels duplicating `NodeInfoLive.@interval_options`.
- 🟡 `lib/voyager_web/components/process_components.ex:156` — `translate_error`/`field_error` duplicate `core_components.ex` because the form bypasses the required `<.input>` component.
- 🟡 `e2e/tests` — no Playwright spec for the new page, while `node_info` and `supervision_tree` have one; none of the crashes above would be caught by the unit suite.
- 🟢 `lib/voyager_web/live/processes_live.ex:280` — the `{:exit, _}` clause does not reset `dirty?`, leaving the table dimmed and inert after a crashed fetch.
- 🟢 `lib/voyager/node_session.ex:44` — `restore_cached_connector_name/1` is a public API used only by a test.
- 🟢 `lib/voyager_web/form_schemas/process_list_controls.ex:63` — clearing the timeout box shows an empty field while the old value is silently used; `validate_required` surfaces it.

<verdict: 3 blocking, 11 should-fix, 3 nit>

---

GuzekAlan:
@claude review opus

lib/voyager_web/live/processes_live.ex
GuzekAlan: Already done — `handle_event("sort", ...)` compares as strings against `Processes.sortable_attrs()` and ignores anything unmatched; `String.to_existing_atom/1` is gone.

lib/voyager_web/live/processes_live.ex
GuzekAlan: Already done — `parse_interval/1` matches the value against `Processes.refresh_interval_options()` and returns `nil` for anything else, which stops the timer instead of scheduling it.

lib/voyager_web/form_schemas/process_list_controls.ex
GuzekAlan: Fixed. The nil guard now lives in the params normalizer rather than in `update_change/3`: a null `search` is dropped so it keeps the current value, instead of being coerced to `""` (which would clear the search when the key is merely absent).

lib/voyager_web/live/processes_live.ex
GuzekAlan: Not doing this one, though the doubled scan is real. Both suggested cures make the first fetch depend on the JS hook running, so a client where the hook fails to load shows a table that never leaves "Scanning processes…" — a worse failure than one extra scan. I tried it: it also breaks 37 tests that mount and assert on rendered rows, and rewriting them to hand-simulate a `pushEvent` would make the suite assert less about real behaviour. Happy to revisit if the second scan shows up as a real cost on a large node.

lib/voyager_web/live/processes_live.ex
GuzekAlan: Not reproducible as described — `sort` does have the in-flight guard. It ends in `fetch/1`, the same function `:auto_refresh`, `refresh_now` and `restore_settings` route through, and nothing calls `start_fetch/2` directly except `mount/3`. Measured it: 48 header clicks during an in-flight scan spawned 2 remote scans (mount plus one after it landed), not 48. If the 24-scan measurement came from a node running a build older than the guard, that would explain it; multiple browser tabs would too, since the guard is per-LiveView.

The related real bug is the one raised at line 316 — the guard dropped the request but kept the assigns — and that is now fixed with a replay.

lib/voyager_web/live/processes_live.ex
GuzekAlan: Already done — the timeout clause sets `AsyncResult.failed` when `page_result.ok?` is false, and `format_error(:timeout)` is the "try a longer timeout or a smaller limit" hint.

lib/voyager_web/live/processes_live.ex
GuzekAlan: Fixed, taking the second option. The "Timeout (ms)" field now carries help text saying the wait is only local and the scan runs to completion on the node. Killing the remote worker via `spawn_request` + `exit(Pid, kill)` is the better fix but it is a change to the agent protocol, not to this page — worth its own PR.

lib/voyager_web/components/core_components.ex
GuzekAlan: Already done — `phx-update="ignore"` is back on the tip, with a comment noting the hook stores open/pinned state and position there and that a tip whose text changes needs a new id to remount.

lib/voyager_web/components/data_table_components.ex
GuzekAlan: Fixed the docstring rather than the component. It no longer claims a stream can be passed: `rows` takes the `{dom_id, row}` shape a stream yields, but the `<tbody>` has no `phx-update="stream"`, so a caller holding a stream renders from a re-fetched list. Adding a real stream mode belongs with the ETS page that will need it, where it can be built against an actual caller.

lib/voyager_web/components/process_components.ex
GuzekAlan: Not doing this now. The count is right and the concern is fair, but both suggested replacements lose something: a `title` attribute cannot be styled or copied from, and a single delegated tooltip is a real rework of the tooltip component that would touch every other page using it. This is a performance problem worth fixing against a measurement on the ETS page, where the row count actually multiplies — not speculatively here.

lib/voyager_web/components/data_table_components.ex
GuzekAlan: Not doing this now. Agreed that hard-coded event names block two tables on one page, but there is exactly one caller today, and the right shape for the parametrisation (per-event attrs vs. a single `target`) depends on whether the ETS page nests its tables in a LiveComponent. Adding attrs now means guessing that, and guessed-at flexibility is the kind that gets rewritten on first real use. It is a mechanical change when the second caller appears.

lib/voyager/queries/processes.ex
GuzekAlan: Not doing this now, but the cycle is real and worth recording. `Voyager.Queries.Processes` matching on a `VoyagerWeb.FormSchemas` struct while that schema calls back is a genuine layering inversion. Unwinding it means changing `page/3` to plain arguments and moving `@refresh_intervals` next to `interval_select`, which touches `NodeInfoLive` too — a refactor with its own review surface rather than a fix folded into this PR. Flagging it so it does not get lost.

lib/voyager_web/components/process_components.ex
GuzekAlan: Half fixed. `translate_error/1` was duplicated byte-for-byte — it turned out to be copied three times (here, `core_components`, and `supervision_tree_live/controls.ex`), so it is now public in `core_components` and both copies are gone.

Kept the raw `<select>`/`<input>` and `field_error/1`. This form is a 3x3 grid — labels row, inputs row, errors row — so the fields line up across columns; `<.input>` wraps label, control and error in one `<div>`, which cannot span three grid rows. Using it would mean giving up the alignment or restructuring the grid around the component.

lib/voyager_web/live/processes_live.ex
GuzekAlan: Already done — the `{:exit, reason}` clause resets `dirty?` like the others. It also clears the queued replay now, since a fetch that crashed would only crash again.

lib/voyager/node_session.ex
GuzekAlan: Already done — `restore_cached_connector_name/1` is marked `@doc false`.

lib/voyager_web/form_schemas/process_list_controls.ex
GuzekAlan: Already done — the changeset has `validate_required([:limit, :timeout])`, and a params normalizer rewrites a blank `limit`/`timeout` to `nil` first, since `cast/4` would otherwise treat `""` as no change and silently keep the old value.

lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,463 @@
...
+    |> assign(:controls, controls)
+    |> assign(:form, to_form(changeset, as: :controls))
+    |> assign(:page, 1)
+  end
+
+  # A scan already in flight is left to finish: `start_async/3` would replace the
+  # task without stopping the work it spawned on the remote, so a burst of
+  # clicks would cost the node a full scan each.
+  defp fetch(socket, priority \\ :high) do
+    if loading?(socket.assigns.page_result), do: socket, else: start_fetch(socket, priority)
```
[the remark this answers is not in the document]
GuzekAlan: Fixed. `fetch/2` now sets `refetch_queued?` when it drops a request, and every terminal `handle_async` clause drains it once the running scan lands — except the failure clauses, which clear it, since the replay would hit the same error. The guard test asserts the replay: click Reductions, click Memory mid-scan, and the second scan goes out with `:memory` when the first returns.

lib/voyager_web/live/processes_live.ex
```diff
@@ -0,0 +1,463 @@
...
+
+  def handle_async(:page_result, {:ok, :skipped}, socket) do
+    socket
+    |> clear_loading()
+    |> noreply()
+  end
+
+  # Rate limiting and timeouts are transient, and the rows already on screen are
+  # still the last good answer, so both flash rather than replacing the table.
+  def handle_async(:page_result, {:ok, {:rate_limited, _retry_after_ms}}, socket) do
```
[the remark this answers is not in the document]
GuzekAlan: Fixed. Both clauses now go through one `transient_error/3`: it flashes over rows that are still the last good answer, and sets `AsyncResult.failed` when `page_result.ok?` is false. Added `format_error(:rate_limited)` and a test that drains the limiter before mount and asserts `#processes-error` renders.

lib/voyager_web/components/core_components.ex
```diff
@@ -437,12 +447,99 @@ defmodule VoyagerWeb.CoreComponents do
...
+        ]}
+      >
+        <span class="grow text-left">{@label}</span>
+        <span class="font-mono text-base-content/60">{@count}</span>
+        <.icon name="icon-chevron-right" class="size-3.5 shrink-0 rotate-90 opacity-60" />
+      </div>
+
+      <div
+        tabindex="0"
+        role="listbox"
```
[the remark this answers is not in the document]
GuzekAlan: Fixed — the panel is `role="group"` with an `aria-label`, and `aria-haspopup="listbox"` is off the trigger. It is a plain checkbox group, which screen readers announce correctly on their own.

assets/js/hooks/table_settings.js
```diff
@@ -0,0 +1,45 @@
...
+  store(settings) {
+    try {
+      localStorage.setItem(this.storageKey, JSON.stringify(settings));
+    } catch (_error) {
+      // Remembering the controls is a convenience, not correctness.
+    }
+  },
+
+  // Always pushed, even with nothing stored: the server holds its first fetch
+  // until this arrives, so staying silent would leave the page empty.
```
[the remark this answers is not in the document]
GuzekAlan: Already done — the comment no longer claims the server holds its first fetch. It now only describes what is still true: the hook pushes stored values up on mount and writes back what the server echoes after validating.

test/voyager_web/live/processes_live_test.exs
```diff
@@ -0,0 +1,854 @@
...
+      refute has_element?(view, "#process-controls fieldset[disabled]")
+
+      # The debounce window itself stays undimmed — the user is still editing.
+      change(view, %{"limit" => "250"})
+      refute render(view) =~ "pointer-events-none select-none opacity-60"
+
+      # The dim spans the control-triggered fetch and clears when it lands.
+      send(view.pid, :refetch)
+      render_async(view)
+      refute render(view) =~ "pointer-events-none select-none opacity-60"
```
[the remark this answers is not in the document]
GuzekAlan: Fixed. The test now blocks the scan on a `receive`, asserts the dim class is present while it is open, then releases it and asserts the dim clears. The `refute`-only version would indeed have passed if the dim never rendered.

test/voyager_web/live/processes_live_test.exs
```diff
@@ -0,0 +1,854 @@
...
+      stub_scan([])
+      {:ok, view, _html} = live(conn, @path)
+      render_async(view)
+      assert_received {:scanned, _args, _timeout}
+
+      # Holds the scan open so the next click lands mid-flight, which is what
+      # `render_click/1` alone cannot produce: it waits for the async to finish.
+      stub(Voyager.ErpcMock, :call, fn _node, _mod, _fun, args, timeout ->
+        send(test, {:scanned, args, timeout})
+        Process.sleep(150)
```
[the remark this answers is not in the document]
GuzekAlan: Fixed — the stub blocks on `receive do :release -> ... end` and the test releases it by pid, so the mid-flight window is deterministic instead of racing 150 ms against `render_async/2`. `Process.sleep/1` is gone from the file.

GuzekAlan:
@claude review opus

hhubert6 (approved):
good job 🐬

PR #210, opened 2026-08-31
Add table records API

piotrsac:
## Summary

Slims [VOY-229](https://linear.app/softwaremansion/issue/VOY-229) (parent [VOY-27](https://linear.app/softwaremansion/issue/VOY-27)): a **LiveView-free** ETS records API that matches process list / process terms — **agent only**.

Rebased onto `main` so `Agent.install`, OTP 27+, and `Agent.call` → `NodeSession.agent_missing/1` on `:undef` are in place.

This is **not** “MFA today, agent tomorrow.” There is no export probe, no `:ets.select` / `:ets.lookup` fallback, no host `Sanitize`, no host `max_heap_size` isolate, and no `Erpc.bind_impl/1`. If the agent is missing, the session is dropped.

The previous dual-path stack (MFA fallback, host sanitizer, isolate, ExUnit live tests) is frozen on [`experimental/ets-mfa-fallback-and-live-tests`](https://github.com/software-mansion/voyager/tree/experimental/ets-mfa-fallback-and-live-tests). Do not recreate it here. Follow-up slims: [VOY-248](https://linear.app/softwaremansion/issue/VOY-248) (230) and [VOY-249](https://linear.app/softwaremansion/issue/VOY-249) (231).

## Do not merge without 230

`Fetch` calls `:voyager_agent.ets_select_chunk/3` and `:ets_lookup/2`. Those exports **do not exist until [VOY-230](https://linear.app/softwaremansion/issue/VOY-230)** (PR [#219](https://github.com/software-mansion/voyager/pull/219)). On main, `Agent.call` treats any `:undef` as “agent gone” and **disconnects**. Merging this PR alone means a peek (or a test against a real node) can drop the whole connection.

Keep the stack: land this with 219 ready to follow (or squash-land them). UI ([VOY-234](https://linear.app/softwaremansion/issue/VOY-234)) may mock or disable fetch until 230.

## Public API

```elixir
# metadata (cheap OTP MFA) — list page, MCP list_ets_tables / get_ets_table_info
Remote.list(node, timeout)
Remote.info(node, table_id, timeout)
TableId.resolve(node, string, ids, timeout)

# records (agent) — contents panel, MCP read_ets_table_chunk
Fetch.select_chunk(node, table_id, limit, continuation \\ nil, timeout)
Fetch.lookup(node, table_id, key, timeout)
```

Chunk is `%{records, continuation}` only (no `via`). Lookup keys: atom / integer / binary. Chunk sizes: 10 / 20 / 50.

`{:remote_exception, :badarg}` → `{:error, :cannot_read}`. A wrapped `{agent_worker_down, {badarg, _}}` is **not** remapped. Nil continuation is sent as `:undefined`. `$end_of_table` / `{records, cont}` are decoded on the host.

Truncation and target heap live on **230**, not here. Until then, records are whatever the (missing) agent would return.

## What was dropped vs the previous #210 body

| Dropped | Why |
|---|---|
| MFA fallback + `probe_export` | Same rule as process terms: missing agent drops the session |
| Host `Sanitize` | Duplicate of 230’s Erlang `truncate_term/1` |
| Host `Fetch` isolate + `@max_heap_size` | MFA survival hatch; forced `bind_impl` because Mox is process-local |
| `Erpc.bind_impl/1` | Only used by ExUnit live tests |
| `*_live_test.exs` | Behaviour coverage moves to Playwright e2e ([VOY-245](https://linear.app/softwaremansion/issue/VOY-245)) |

## Agent contract (lock for VOY-230)

```
:voyager_agent.ets_select_chunk(Table, Limit, Continuation | undefined)
  -> '$end_of_table' | {Records :: list(), Continuation}
:voyager_agent.ets_lookup(Table, Key) -> list()
```

A non-`undefined` continuation has crossed ETF; `ets_select_chunk/3` **must** call `ets:repair_continuation/2` on the target before `ets:select/1`. A one-shot worker’s `badarg` **must** be re-raised as `error:badarg` in the erpc process so this facade can map it to `:cannot_read`.

## Test plan

- [x] `mix test test/voyager/services/ets/ test/voyager/erpc_test.exs`
- [x] `mix precommit`
- [ ] Invalid limit / table / key rejected locally
- [ ] Decode `$end_of_table` / `{records, cont}` / lookup list
- [ ] `badarg` → `:cannot_read`; wrapped worker death is not
- [ ] Missing export is **not** retried as `:ets.select` / `:ets.lookup`
- [ ] List/info still do not call `:voyager_agent`


piotrsac:
@claude review opus

piotrsac:
@claude review opus

piotrsac:
@claude review opus

piotrsac:
@claude review

piotrsac:
@claude review opus

lib/voyager/services/ets/fetch.ex
```diff
@@ -0,0 +1,80 @@
...
+          {:ok, chunk()} | {:error, term()}
+  def select_chunk(node, table, limit, continuation \\ nil, timeout \\ Erpc.default_timeout()) do
+    isolated_read(timeout, fn ->
+      Remote.select_chunk(node, table, limit, continuation, timeout)
+    end)
+  end
+
+  @spec lookup(node(), TableId.t(), atom() | integer() | binary(), timeout()) ::
+          {:ok, chunk()} | {:error, term()}
+  def lookup(node, table, key, timeout \\ Erpc.default_timeout()) do
```
GuzekAlan: 🟡 should-fix: `lookup/4` takes no limit and applies none, but on a `bag` / `duplicate_bag` one key legitimately holds N objects. Measured against a peer node: a `duplicate_bag` key with 3 000 copies returns all 3 000 records (the caller asked for a 10/20/50 chunk), and at 100 000 copies the row is permanently unreadable — `{:error, :heap_limit_exceeded}` every time, with the target having copied ~40 MB first. `select_chunk/5` enforces `limit in [10, 20, 50]`; `lookup/4` should take the same limit and slice.

---

lib/voyager/services/ets/remote.ex
```diff
@@ -71,6 +91,122 @@ defmodule Voyager.Services.Ets.Remote do
...
+      {:ok, result} -> decode_lookup(result, :mfa)
+      {:error, _} = err -> map_read_error(err)
+    end
+  end
+
+  defp decode_select(:"$end_of_table", via) do
+    {:ok, %{records: [], continuation: nil, via: via}}
+  end
+
+  defp decode_select({records, continuation}, via) when is_list(records) do
```
GuzekAlan: 🔴 blocking: when a page exhausts the table `:ets.select/3` returns `{Objects, :"$end_of_table"}`, so this clause hands back `continuation: :"$end_of_table"` instead of `nil` — the clause above establishes `nil` as "no more rows", and every table smaller than the chunk size hits this. Verified against a peer node: 3 rows at limit 10 gives `{3, :"$end_of_table"}`, and feeding that back is `{:error, :cannot_page}` on MFA and a wasted probe+read round trip returning an empty page on the agent path. Add a `decode_select({records, :"$end_of_table"}, via)` clause mapping the continuation to `nil`.

---

lib/voyager/services/ets/fetch.ex
```diff
@@ -0,0 +1,80 @@
...
+    %{chunk | records: Enum.map(records, &Sanitize.term/1)}
+  end
+
+  defp isolate(timeout, fun) do
+    impl = Erpc.impl()
+
+    task =
+      Task.Supervisor.async_nolink(Voyager.TaskSupervisor, fn ->
+        Erpc.bind_impl(impl)
+        Process.flag(:max_heap_size, %{size: @max_heap_size, kill: true})
```
GuzekAlan: 🟡 should-fix: `max_heap_size` does not account refc binaries, so this cap does not bound the payload the moduledoc says it bounds. Measured on a peer node: one row holding a 200 MB binary took the host VM from 56 MB to a 246 MB peak over 1.5 s and *then* returned a 512-byte prefix — `include_shared_binaries: true` does not change it. At `limit: 50` over a table of large binaries this is an OOM path for Voyager, so either bound it another way or narrow the "Protects Voyager" claim to non-binary payloads.

---

lib/voyager/services/ets/remote.ex
```diff
@@ -71,6 +91,122 @@ defmodule Voyager.Services.Ets.Remote do
...
+
+  defp decode_select(_other, _via), do: {:error, :invalid_response}
+
+  defp decode_lookup(records, via) when is_list(records) do
+    {:ok, %{records: records, continuation: nil, via: via}}
+  end
+
+  defp decode_lookup(_other, _via), do: {:error, :invalid_response}
+
+  defp map_read_error({:error, {:remote_exception, :badarg}}), do: {:error, :cannot_read}
```
GuzekAlan: 🟡 should-fix: this only recognises the shape MFA produces. The agent contract in the PR body requires `ets_select_chunk/3` to run in a spawned one-shot worker, and a `badarg` raised there does not reach `safe_call/5` as `{:remote_exception, :badarg}` unless the agent re-raises it in the erpc process. Confirmed with a contract-compliant agent on a peer node: a private table came back as `{:error, {:remote_exception, {:agent_worker_down, {:badarg, _}}}}`, so `:cannot_read` is lost and VOY-234's "private table → cannot read" rule breaks on the agent path. Document the re-raise requirement alongside the `repair_continuation/2` one.

---

lib/voyager/services/ets/remote.ex
```diff
@@ -71,6 +91,122 @@ defmodule Voyager.Services.Ets.Remote do
...
+  defp map_read_error({:error, _} = err), do: err
+
+  defp probe_export(node, fun, arity, timeout) do
```
GuzekAlan: 🟡 should-fix: the probe is private and its result only surfaces as `chunk.via`, which the caller learns *after* the target has already copied full objects. VOY-234's "must not skip" list requires refusing the fetch when the agent export is missing rather than silently MFA-ing, and that decision has to be made before the read. Export a public `agent?/2` (or return the probe result from a cheap call) so the LiveView can disable Fetch instead of discovering `via: :mfa` in the response.

---

test/voyager/services/ets/remote_live_test.exs
```diff
@@ -67,6 +64,88 @@ defmodule Voyager.Services.Ets.RemoteLiveTest do
...
+
+  test "select_chunk/3 and lookup/3 use an unnamed table reference as the handle" do
+    tid = :ets.new(unique_name(), [:public])
+    on_exit(fn -> safe_delete(tid) end)
+    :ets.insert(tid, {:k, 1})
+
+    assert is_reference(tid)
+    assert {:ok, %{records: [{:k, 1}]}} = Remote.lookup(Node.self(), tid, :k)
+
+    assert {:ok, %{records: [{:k, 1}], via: :mfa}} =
```
GuzekAlan: 🟡 should-fix: this is the only case in the suite where the table is smaller than the limit (1 row, limit 10), and it is the case that returns `continuation: :"$end_of_table"` — but the assertion only matches `records` and `via`, so the defect passes. Assert `continuation: nil` here; the other select tests all use 25 rows at limit 10 and never reach the end of a page.

---

GuzekAlan (commented):
## 🦀 Review

- 🔴 `lib/voyager/services/ets/remote.ex:183` — `:ets.select/3` returns `{Objects, :"$end_of_table"}` on the page that exhausts the table, so `continuation` comes back as `:"$end_of_table"` instead of `nil`. Any table smaller than the chunk size is affected; the caller cannot tell it is done, and feeding the value back is `:cannot_page` on MFA / an empty page on the agent path. Add a `decode_select({records, :"$end_of_table"}, via)` clause.
- 🟡 `lib/voyager/services/ets/fetch.ex:32` — `lookup/4` applies no record cap. A `duplicate_bag` key with 3 000 copies returns all 3 000 records; at 100 000 the row is permanently `:heap_limit_exceeded`. Take the same 10/20/50 limit as `select_chunk/5`.
- 🟡 `lib/voyager/services/ets/fetch.ex:57` — `max_heap_size` does not account refc binaries, so the cap does not bound binary payloads. One 200 MB binary row moved the host VM from 56 MB to a 246 MB peak before being cut to 512 bytes; `limit: 50` multiplies that.
- 🟡 `lib/voyager/services/ets/remote.ex:195` — `map_read_error/1` only matches the MFA `badarg` shape. The agent contract mandates a spawned worker, whose `badarg` does not arrive as `{:remote_exception, :badarg}` unless re-raised in the erpc process, so `:cannot_read` is lost on the agent path.
- 🟡 `lib/voyager/services/ets/remote.ex:198` — the export probe is private and `via` is only known after the target has copied the objects, so VOY-234 cannot refuse an MFA fetch before it happens.
- 🟡 `test/voyager/services/ets/remote_live_test.exs:137` — the one test with fewer rows than the limit does not assert on `continuation`, so the defect above passes.
- 🟢 `lib/voyager/erpc.ex:5` — the `@moduledoc` still says the implementation is resolved from application config; `bind_impl/1` now wins over it.

Table-option matrix (`set` / `ordered_set` / `bag` / `duplicate_bag` × `public` / `protected` / `private`, named and unnamed, `keypos` 1–3, `compressed`, `read_concurrency`, `write_concurrency` `true|false|:auto`, `decentralized_counters`, `heir`) reads correctly through `info/3` and `list/2` on a real peer node, and a full agent-path drain of each type returned every row exactly once. The sanitizer survived 20 000 random terms with no crash, no non-idempotent output and no cap violation.

<verdict: 1 blocking, 5 should-fix, 1 nit>

---

lib/voyager/services/ets/fetch.ex
piotrsac: Valid measurement, out of scope for this ticket.

`lookup/4` is OTP `:ets.lookup/2` (all objects for the key); `select_chunk/5` is the 10/20/50 pager. A host-side slice would not stop the target copy or the 100k `heap_limit_exceeded`. [VOY-230](<https://linear.app/softwaremansion/issue/VOY-230>) truncates **terms** (512/50/5) on `ets_lookup/2`, it does not add a limit.

Bounding a bag key needs a key match spec + Limit on the target (new export or VOY-231-style select), not `Enum.take` in Fetch. Tracked as [VOY-241](<https://linear.app/softwaremansion/issue/VOY-241/ets-bounded-bag-key-lookup>).

lib/voyager/services/ets/remote.ex
piotrsac: Wrong ticket, and the must-not-skip citation is wrong.

This is [VOY-229](<https://linear.app/softwaremansion/issue/VOY-229>): missing exports → MFA, `via: :mfa` on the chunk. No LiveView. [VOY-230](<https://linear.app/softwaremansion/issue/VOY-230>) says until [VOY-210](<https://linear.app/softwaremansion/issue/VOY-210>) the exports do not exist on real nodes and this API stays on MFA.

[VOY-234](<https://linear.app/softwaremansion/issue/VOY-234>)'s must-not-skip list is: drive the banner from `chunk.via`; do not MFA after heap kill / timeout / noconnection / `undef`; do not inject. Refusing a missing-export fetch is Outcome copy for the LiveView ("disable the button"), not a must-not-skip item, and not this PR. The post-failure MFA retry is already covered (undef after a true probe does not fall back).

A public `agent?/2` is a VOY-234 convenience. Changing `Remote` to refuse MFA would break this ticket.

test/voyager/services/ets/remote_live_test.exs
piotrsac: Valid. The map pattern only matched `records` and `via`, so `continuation: :"$end_of_table"` still passed.

Done in `2b95f05`: this live assert now includes `continuation: nil`, and unit tests cover the `{records, :"$end_of_table"}` mapping on both MFA and agent paths.

lib/voyager/services/ets/fetch.ex
```diff
@@ -0,0 +1,91 @@
...
+  Runs `Remote.select_chunk/5` and `Remote.lookup/4` in a TaskSupervisor child
+  with `max_heap_size` 500_000 words (`kill: true` and `include_shared_binaries:
+  true`), then sanitizes records (not the continuation). Heap kill is
+  `{:error, :heap_limit_exceeded}`; a wait that expires is `{:error, :timeout}`.
+
+  The cap is this task's process heap (cons cells, maps, tuples) plus off-heap
+  binaries it refers to, not host RSS. MFA peek still copies full objects on
+  the remote node and over the wire; distribution allocates them before the
+  task can be killed, then Sanitize keeps a 512-byte binary prefix. A page of
+  large binaries can still OOM Voyager.
```
GuzekAlan: Can you shorten the mdouledoc and add only essential info

lib/voyager/erpc.ex
```diff
@@ -30,10 +30,22 @@ defmodule Voyager.Erpc do
...
+  end
+
+  @doc false
+  @spec bind_impl(module()) :: :ok
+  def bind_impl(module) when is_atom(module) do
+    Process.put(impl_key(), module)
+    :ok
+  end
+
+  defp impl_key, do: {__MODULE__, :impl}
```
GuzekAlan: Isn't it overkill? What are live tests (LiveVIewTest or E2E) ?

lib/voyager/services/ets/remote.ex
```diff
@@ -71,6 +91,133 @@ defmodule Voyager.Services.Ets.Remote do
...
+      {:error, _} = err -> map_read_error(err)
+    end
+  end
+
+  defp mfa_lookup(node, table, key, timeout) do
+    case Erpc.safe_call(node, :ets, :lookup, [table, key], timeout) do
+      {:ok, result} -> decode_lookup(result, :mfa)
+      {:error, _} = err -> map_read_error(err)
+    end
+  end
```
GuzekAlan: If there is `agent` implementation then leave use them. Our logic is to drop connection if there is no `voyager_agent` on remote

lib/voyager/services/ets/remote.ex
```diff
@@ -71,6 +91,133 @@ defmodule Voyager.Services.Ets.Remote do
...
+       when not is_nil(continuation) do
+    {:error, :cannot_page}
+  end
+
+  defp mfa_select(node, table, limit, nil, timeout) do
+    case Erpc.safe_call(node, :ets, :select, [table, @match_all, limit], timeout) do
+      {:ok, decoded} -> decode_select(decoded, :mfa)
+      {:error, _} = err -> map_read_error(err)
+    end
+  end
```
GuzekAlan: Same as before. 2 implementations seem like overkill. If agent implementation will be added later then just assume it will be added

lib/voyager/services/ets/sanitize.ex
```diff
@@ -0,0 +1,183 @@
+defmodule Voyager.Services.Ets.Sanitize do
```
GuzekAlan: @hhubert6 wouldn't this module be useful in both process and ets ? It seems like general tool for elixir terms. It might be `Voyager.Services.Term.Sanitize`

lib/voyager/services/ets/fetch.ex
```diff
@@ -0,0 +1,91 @@
...
+  """
+
+  alias Voyager.Erpc
+  alias Voyager.Services.Ets.Remote
+  alias Voyager.Services.Ets.Sanitize
+  alias Voyager.Services.Ets.TableId
+
+  @max_heap_size 500_000
```
GuzekAlan: Please add comment how large it is in MB (words are hard to think of). It should be around 100MB. This is Voyager max_heap_size so it can be large. max_heap_size on remote should be smaller

lib/voyager/erpc.ex
piotrsac: Not LiveView and not E2E — `fetch_live_test` / `remote_live_test` hit real `:erpc` and real ETS on `Node.self()`, versus the Mox tests in `fetch_test`.

`test_helper` pins `:erpc` to `ErpcMock` for the whole VM. `async: false` still overlaps `async: true` Mox tests, so `Application.put_env` would race (same pattern as `node_info_test`). `bind_impl` is process-local to avoid that.

`Fetch.isolate/2` re-binds into the Task because the process dictionary is not inherited. That's the cost; `put_env` would be simpler and flaky. I'll reword the `@doc` so it doesn't say "live tests".

lib/voyager/services/ets/sanitize.ex
piotrsac: Yes — it's a general term cap, not ETS-shaped. Leaving it here for this PR because 229's spec is `Ets.Sanitize` and 230 copies the same 512/50/5 contract into the agent.

Process already has a different walk: VOY-187 `bound_term` (visit budget + bare `:'$voyager_truncated'`). Unifying those blindly is what 230/242 warn against.

Follow-up: lift the Elixir module to `Voyager.Services.Term.Sanitize`, keep the structured markers as the inspector contract, and take both policies (`:structural` for ETS pages, `{:budget, n}` for process state/mailbox/dictionary). Ticket: https://linear.app/softwaremansion/issue/VOY-246/shared-term-sanitizer-termsanitize

hhubert6: With assumption of sanitization/truncation logic being always present on remote node having it also directly in voyagers seems like overkill (at least for now).

piotrsac:
@claude review

lib/voyager/services/ets/fetch.ex
hhubert6, line 1: This API should allow setting size of fetched data. For example as it done in process info with budget.

hhubert6 (changes requested):

lib/voyager/services/ets/fetch.ex
piotrsac: #219

PR #215, opened 2026-09-02
Chore: Default the PR review workflow to Opus

dazakdev:
## What

`@claude review` now runs the review with `--model opus` by default. `@claude review sonnet` opts into Sonnet.

## Why

Opus gives noticeably better review findings; Sonnet was the default only as a cost saving. Flipping the default keeps the cheaper option available on demand.

## Notes for reviewers

The old `@claude review opus` phrase still triggers a review (it contains `@claude review`) and now simply gets Opus like everything else. No other file referenced the old trigger.


piotrsac (approved):
LGTM 🦭

PR #216, opened 2026-09-02
Feature: ETS table list page

GuzekAlan:
Closes [VOY-228](https://linear.app/softwaremansion/issue/VOY-228/ets-table-list-page). Rebased onto `main` after #208 (process list page) merged; it does not modify any of the process page's own files.

## What

`/node/:node/ets-tables` lists every ETS table of the connected node through `Voyager.Services.Ets.Remote` — plain OTP MFAs, no injected code, no mocks — replacing the coming-soon placeholder. Built as a sibling of the process list page:

- **Fetcher, same lifecycle as the process page's.** First fetch waits for the client's stored controls (one fetch per visit), control changes refetch through the same 1.5s debounce, mid-flight requests queue and replay. Auto-refresh Off/5s/10s/30s/60s, default 5s, ticks at `:low` rate-limiter priority.
- **One fetch, everything local.** `EtsTablesLive.Query.all/2` wraps `Remote.list/2`; search, filters, sort and paging run on the last result. Sortable columns: Table, Objects, Memory, Keypos. `EtsTableListControls` mirrors `ProcessListControls`, deriving its column sets from `Query`.
- **Filter selects for Protection, Type and Named** next to the free-text search — applied locally like the search, stored with the other controls, and a stale stored pick degrades to Any.
- **Column picker with every extractable attribute**: protection, type, objects, owner, named, keypos, heir, compressed, read/write concurrency, decentralized counters. Table and Memory locked; protection/type/objects/owner on by default. Headers always show their full label; a wide column set scrolls inside the table's card behind a minimum width derived from the columns, and the card is the page's only scroll container on both axes.
- **Side panel with the full metadata.** `?table=` selection re-resolved against every fetch; the panel shows everything the list fetch already carries (Overview + Storage), so there is no separate details page or second fetch. A truncated name's tooltip carries the full name and, for an unnamed table, its reference — both copyable.

## Footprint on the process page's code

None — `ProcessComponents`, `ProcessListControls`, `ProcessesLive` and friends are byte-identical to `main`. Shared modules only gain additive pieces the ETS page renders through: `DataTableComponents` (`value_cell`, round-trip label, `row_class`, `:xs`/`:lg` widths, derived min-width), `DetailsPanelComponents` (`copyable` public), sidebar/router entries. The ETS page adopts the same full-height layout wrappers the process page gained on `main`.

## Tests

- Unit tests cover the local logic — `Query` (sort/filter/fetch parsing over stubbed raw `:ets` replies via `Voyager.EtsFakes`), `EtsTableListControls`, and the `DataTableComponents` additions. The page itself is covered end to end, matching the process list page.
- **Playwright e2e** (`e2e/tests/ets_tables.spec.ts`, Chromium + Firefox): list + summary, search, filter selects, local sort, column picker (hide + reveal), panel with full metadata, private table, unknown table — against the real fixture node; `mock_app` owns four known tables (`mock_ets_owner`). Local runs need `mix assets.build` first (CI's `assets.deploy` covers it).

## Checks

`mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict`, `mix dialyzer` and `mix test` pass, and the Playwright suite passes in Chromium and Firefox. The filters, tooltips, header labels and single-scroll layout were verified in a real browser against the fixture node.

https://claude.ai/code/session_0165zRaMa9mncSWtaTGLauxS


GuzekAlan:
@claude review opus

piotrsac:
Screenshot 2026-09-08 at 14 46 18
there is a situation where before a sidebar is collapsed, the view of the table is scrollable horrizontally, i think it would be better to collapse the sidebar quicker, before that to ensure better view and reduce number of situations where we should scroll horizontally

piotrsac:
LGTM 🦭

piotrsac (approved):

PR #221, opened 2026-09-03
Feature: Add term inspector

srzeszut:


srzeszut:
@claude review opus

srzeszut:
@claude review opus

srzeszut:
@claude review opus

srzeszut:
@claude review

lib/voyager_web/term_tree.ex
```diff
@@ -0,0 +1,413 @@
...
+
+  def children(list, offset, limit) when is_list(list) do
+    case list_length(list) do
+      :improper ->
+        []
+
+      {:ok, _count} ->
+        slice = Enum.slice(list, offset, limit)
+
+        if Keyword.keyword?(list) do
```
GuzekAlan, line 79: Keyword is visible differently when truncated and not. Quickest fix might be checking if last element is truncate term and changing it to Keyword truncate term (e.g. {truncate_term, true}). But it can wait since it is visual bug. So it might be better to create an issue on it and merge it as fast as possible

Image

GuzekAlan (commented):
I've got this error when inspecting improper list 

Image

GuzekAlan (approved):
Apart from error which wasn't reproduced looks ok. If you cannot find the error with improper lists let's merge it and see how it works

hhubert6 (approved):
Excellent 🐬

PR #223, opened 2026-09-03
Add installing voyager_agent on node connect

hhubert6:
Connect now owns loading the agent: after a successful distribution connect, NodeSession loads :voyager_agent on the remote node and calls `register/1`. If that fails, the connect fails cleanly — no half-connected session.

- `Agent.install/1`: OTP-27 check first (older ⇒ connect refused with an explicit version message, no code sent), then load, then register(Node.self()).
- `Agent.call/4` drops the session when a remote call returns `:undef` (agent gone mid-connection).
- Error messages in both connect forms; `agent_install_failed` allowlisted in telemetry.

### DIFF
--- a/lib/voyager/agent.ex
+++ b/lib/voyager/agent.ex
@@ -22,6 +22,10 @@ defmodule Voyager.Agent do
            | {:register_failed, term()}
            | CodeInjector.error_reason()}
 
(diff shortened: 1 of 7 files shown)


hhubert6:
@claude review

hhubert6:
@claude review

GuzekAlan (approved):
🐙

PR #224, opened 2026-09-03
Feat: Process info page

GuzekAlan:


hhubert6:
- When navigating to process info sidebar auto opens even if closed.
- Pids in overview could be links as well
- Pid of inspected process should take space right of the node name IMO.
- Term fetching should not have upper limit of budget input.
- Messages, dictionary, links, monitors and monitored_by also have number limit as input so it should be included in users parameters.
- As rate limiter provides it the time it took to fetch data also could be displayed (it’s valuable info).

GuzekAlan:
All six points are addressed as of 9d94b3e:

- **Sidebar auto-opening** — fixed in 3c2ef0c: the `?sidebar=` mode is now carried onto the Show More link, the relations chips and the back link via a `keep_sidebar/2` helper. Note the process *list* page's links live on main, so they'll need the same one-liner once the branches meet.
- **Pids in overview as links** — Parent and Group leader now link to their process pages when they belong to the inspected node (9d94b3e).
- **Pid right of the node name** — done in c19fdf3 (moved into the node header's actions slot).
- **No upper limit on budget** — the input keeps only its minimum; the remote-side truncation is the actual safety (9d94b3e).
- **Limits as inputs** — Relations, Messages and Dictionary each have a Limit control next to Budget/Timeout, wired through to the fetches (c19fdf3).
- **Fetch duration** — each section shows "fetched HH:MM:SS UTC in N ms", using the rate limiter's own measurement (9d94b3e).

Also fixed in passing: the flaky `export_test` failure on CI was a leaked `:fake` connector name in the `persistent_term` cache that `node_session_test`'s teardown didn't restore (7a06473).

---

hhubert6:
- It'd be nice if fetch time have colors the same way as in process list
- PID in top right corner should have larger font IMO

GuzekAlan:
Both done in b576ca8: fetch durations use the process list's slow-fetch colors (red past 3 s, warning past 1 s), and the header pid matches the node name's size.

---

PR #239, opened 2026-09-09
ETS bounded bag key lookup

piotrsac:
## Summary

`ets:lookup/2` returns every object for a key. On `:bag` / `:duplicate_bag` that can copy tens of MB onto the target and kill the worker.

This replaces it with a keyed `ets:select` (hash lookup, not a scan), capped by Limit + continuation. Search `key_eq` uses the same path. Contents-view Lookup stays set/ordered_set until [VOY-259](https://linear.app/softwaremansion/issue/VOY-259).


piotrsac:
@claude review opus

piotrsac:
@claude review

priv/voyager_agent.erl
```diff
@@ -528,12 +530,88 @@ ets_select_spec(Table, Spec, Limit, Budget, Cont)
...
+                false ->
+                    keyed_clauses(Keypos, Key)
+            end;
+        _ ->
+            erlang:error(badarg)
+    end.
+
+keyed_clauses(Keypos, Key) ->
+    [{erlang:setelement(Keypos, erlang:make_tuple(N, '_'), Key), [], ['$_']}
+     || N <- lists:seq(Keypos, ?ETS_LOOKUP_MAX_ARITY)].
```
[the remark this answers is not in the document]
piotrsac: someone who makes a table with `keypos: 300` deserves that

piotrsac:
@claude review

piotrsac:
@claude review

piotrsac:
@claude review

priv/voyager_agent.erl
```diff
@@ -528,12 +530,88 @@ ets_select_spec(Table, Spec, Limit, Budget, Cont)
...
+            erlang:error(badarg)
+    end;
+do_lookup(Table, Key, _Limit, Budget, Cont) ->
+    wrap_select(ets:select(
+                    ets:repair_continuation(Cont, lookup_spec(Table, Key))),
+                Budget).
+
+bag_lookup(Table, Key, Limit, Budget) ->
+    Spec = lookup_spec(Table, Key),
+    case ets:select(Table, Spec, Limit) of
```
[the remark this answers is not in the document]
piotrsac: Agreed on the invariant: leftover objects must not leave the target. MCP `key_eq` would otherwise ship an unbounded cursor and blow `EtsHelpers`' `@max_cursor_bytes`.

Capping or stripping that leftover list in place is not a fix. After a bound-key select, `slot_ix` is already `-1`, so the leftover list is the only resume token (`assert_paged_key_lookup/1` depends on it). Emptying it drops the rest of the key; running `bound_term/2` over those terms would re-emit the truncated copies as the next page.

Lookup now returns `{'$voyager_skip', Skip}` and rebuilds the page on the target (`ets:select/3` then local `ets:select/1`, leftover dropped before the reply). Unbound `ets_select_chunk` / `ets_select_spec` still forward hash leftovers — same class of bug for a fat bag bucket during a table peek, follow-up rather than this lookup path.

piotrsac:
@claude review

priv/voyager_agent.erl
```diff
@@ -528,12 +532,142 @@ ets_select_spec(Table, Spec, Limit, Budget, Cont)
...
+    undefined;
+skip_token(Seen, true) ->
+    {?SKIP_CONT, Seen}.
+
+drop(N, List) when N =< 0 ->
+    List;
+drop(_N, []) ->
+    [];
+drop(N, [_ | Rest]) ->
+    drop(N - 1, Rest).
```
hhubert6: Maybe simplify like this?

```suggestion
bag_page(Table, Key, Limit, Budget, Skip) ->
    Want = Skip + Limit + 1,
    case collect(ets:select(Table, lookup_spec(Table, Key), Limit), Want, []) of
        [] ->
            %% Keyed heads stop at arity 255; lookup still hashes a wider row.
            page(ets:lookup(Table, Key), Skip, Limit, Budget);
        Rows ->
            page(Rows, Skip, Limit, Budget)
    end.

page(Rows, Skip, Limit, Budget) ->
    Page = lists:sublist(lists:nthtail(min(Skip, length(Rows)), Rows), Limit),
    Taken = Skip + length(Page),
    wrap_records(Page, skip_token(Taken, length(Rows) > Taken), Budget).

%% ets:select/3 may return fewer rows than asked with a live continuation.
collect('$end_of_table', _Want, Acc) ->
    lists:reverse(Acc);
collect({Rows, Cont}, Want, Acc) ->
    Acc1 = lists:reverse(Rows, Acc),
    case length(Acc1) >= Want of
        true ->
            lists:reverse(Acc1);
        false ->
            collect(ets:select(Cont), Want, Acc1)
    end.

skip_token(_Taken, false) ->
    undefined;
skip_token(Taken, true) ->
    {?SKIP_CONT, Taken}.
```

priv/voyager_agent.erl
```diff
@@ -528,12 +532,142 @@ ets_select_spec(Table, Spec, Limit, Budget, Cont)
...
+    false;
+ms_pattern_tuple(Key, Index, Size) ->
+    ms_pattern_key(element(Index, Key)) orelse ms_pattern_tuple(Key, Index + 1, Size).
+
+ms_pattern_list([]) ->
+    false;
+ms_pattern_list([Head | Tail]) ->
+    ms_pattern_key(Head) orelse ms_pattern_list(Tail);
+ms_pattern_list(Tail) ->
+    ms_pattern_key(Tail).
```
hhubert6: Maybe shorten like this?

```suggestion
%% Any '$'-prefixed atom counts, not just a real '$N' variable: the cost of a false positive is the slower guard spec.
ms_pattern_key(Key) when is_map(Key) ->
    true;
ms_pattern_key(Key) when is_tuple(Key) ->
    ms_pattern_key(tuple_to_list(Key));
ms_pattern_key([Head | Tail]) ->
    ms_pattern_key(Head) orelse ms_pattern_key(Tail);
ms_pattern_key(Key) when is_atom(Key) ->
    case atom_to_list(Key) of
        [$$ | _] ->
            true;
        "_" ->
            true;
        _ ->
            false
    end;
ms_pattern_key(_) ->
    false.
```

hhubert6 (commented):
Looking good

priv/voyager_agent.erl
piotrsac: chose entirely different approach, deleting most of that code in the end, resigning from paging

priv/voyager_agent.erl
piotrsac: changed approach for counting size only

PR #245, opened 2026-09-14
Feature: Add help tooltips to process and ETS panels and a connect tutorial

dazakdev:
Closes VOY-84

## What

- Process overview and memory/GC rows (supervision tree side panel and process info page) and the ETS table details panel get the same "?" help affordance as the node info page. Each entry is a short plain-language explanation plus a link to the Erlang docs.
- Help text lives in `VoyagerWeb.ProcessInfoHelp` and `VoyagerWeb.EtsTableHelp`, same shape as `NodeInfoHelp`. The shared `kv/1` row component accepts a `help` entry, so all three panels pick it up from one change.
- The ETS tables list shows column help on hovering the header label. `DataTableComponents.table/1` takes an optional `help` entry per column; ETS columns read it from `EtsTableHelp`. The records info panel on the ETS table page gets the same help per field.
- Links, Monitors and Monitored by on the process info Relations tab, and Links in the supervision tree side panel, show help taken straight from the supervision tree legend (`SupervisionTreeComponents.edge_legend/1`), so the wording stays in one place.
- Loading skeleton rows render the same help tooltips as the loaded rows, so rows keep their height when the data lands.
- New `docs/connecting_to_a_node.md` guide: direct and SSH connections, what each form field means, and a "when it does not connect" checklist. The connect page links to it under the heading, with the help icon after the text. The link is grayed out while connected or connecting.
- `link_tooltip/1` forwards `pinnable`, so clicking a header label sorts without pinning its tooltip.

Stacked on top: #248 (process table column help, list page layout fixes, copyable ETS protection).

## Tests

- New `details_panel_components_test.exs` covers the `kv` help tooltip, help in loading skeletons, the relation help on `section/1`, help entries for every ETS column, and the ETS panel rows.
- `data_table_components_test.exs` asserts a column with a help entry renders its header tooltip.
- `connect_live_test.exs` asserts the tutorial link and that it is grayed out while an SSH connection is in progress.
- Four assertions in `details_panel_test.exs` now match the status value cell instead of panel text, because the status tooltip text contains "running" and "waiting".


dazakdev:
@claude review

docs/connecting_to_a_node.md
```diff
@@ -0,0 +1,72 @@
...
+
+- **"Node down" right away**: the node is not distributed, or the name type toggle does not match. Run `node()` in the node's shell; `:nonode@nohost` means no name was given at boot.
+- **Connection refused / timeout**: the host or the port is not reachable. For a direct connection check that `epmd` and the distribution port (random by default, see `inet_dist_listen_min` / `inet_dist_listen_max`) are open in the firewall. For SSH check that the SSH host itself can reach the node.
+- **"SSH authentication failed"**: with SSH Agent, check that `ssh-add -l` lists a key the host accepts and that the agent is running. If it lists several keys of the same type, only the first is tried; see the Authentication note above. With Password, check the password. Try `ssh user@host` from a terminal first, keeping in mind that plain `ssh` tries every key and Voyager does not.
+- **Cookie mismatch**: the node logs `** Connection attempt from disallowed node ... **`. Compare the cookie with `:erlang.get_cookie()` on the node.
+- **Works from the shell, not from Voyager**: make sure Voyager's own name type (Settings → Distribution) matches the node's. Mixed short and long names never connect.
```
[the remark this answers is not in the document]
dazakdev: fixed

dazakdev:
@claude review

lib/voyager_web/process_info_help.ex
```diff
@@ -0,0 +1,164 @@
...
+  @moduledoc """
+  Static help/tooltip metadata for the process overview and memory rows.
+
+  Same entry shape as `VoyagerWeb.NodeInfoHelp`; entries are keyed by a stable
+  atom and passed straight to the `help` attr of `kv/1`.
+  """
+
+  @type entry :: VoyagerWeb.NodeInfoHelp.entry()
+
+  @erts "https://www.erlang.org/doc/apps/erts/erlang.html"
```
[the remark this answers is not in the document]
dazakdev: Leaving as is. It is a third copy (supervision_tree_components.ex has the same literal); a shared module for one docs URL adds indirection without preventing anything real, and a grep for erlang.org/doc finds all of them if the path ever changes.

lib/voyager_web/live/connect_live.ex
```diff
@@ -38,9 +38,18 @@ defmodule VoyagerWeb.ConnectLive do
...
               </h1>
+              <.link
+                id="connect-tutorial-link"
+                href="https://github.com/software-mansion/voyager/blob/main/docs/connecting_to_a_node.md"
+                target="_blank"
+                rel="noopener"
+                class="text-primary mb-4 inline-flex items-center gap-1 text-xs hover:underline"
+              >
+                <.icon name="icon-circle-help" class="size-3.5" /> How to prepare your node
+              </.link>
```
piotrsac, line 69: I think the "How to prepare your node" should go gray while you are already connected to a node

how would circle-help icon look on the other side of the text?

lib/voyager_web/components/details_panel_components.ex
piotrsac, line 1: In the details sidepanel in supervision tree we currently do a weird thing - while we fetch data there are labels, but there are no help tooltips, and there are dark spots in the places where data go - its not that much consistent with other places where we fetch data (node info, supervision tree where there is nothing while we dont have fetched data) - i would consider what to do with it - and especially regarding this PR - should we really not render the help buttons while the data is not fetched?
(especially because the height of the rows change between the fetch and waiting for a fetch)

lib/voyager_web/process_info_help.ex
```diff
@@ -0,0 +1,164 @@
...
+      text:
+        "Token used by the seq_trace facility to follow a message flow across processes. Empty unless seq_trace is active.",
+      doc_href: "https://www.erlang.org/doc/apps/kernel/seq_trace.html",
+      doc_label: "See seq_trace"
+    },
+    error_handler: %{
+      text:
+        "Module called when the process calls an undefined function. Almost always error_handler.",
+      doc_href: @erts <> "#process_flag_error_handler",
+      doc_label: "See process_flag(error_handler)"
```
piotrsac: im not sure about if process_flag/2 is the right function to refer to here, that function does not explain what that is, its just a function to change the error handler module - i would consider putting maybe the most common, default `https://www.erlang.org/doc/apps/kernel/error_handler.html` here, because i dont feel like that function belongs here

@srzeszut i would like to hear your opinion on this one

lib/voyager_web/components/details_panel_components.ex
piotrsac, line 1: im not sure if exactly in this file, but in the for example process details view -> Relations there should be tooltips explaining what is link, monitored by and monitors, preferably the same as used in the supervision tree legend

piotrsac (commented):
I feel like we are missing coverage for the whole ETS list page and ETS table records page view, i think its in this issue scope to cover tooltips there (for example table headers)

docs/connecting_to_a_node.md
```diff
@@ -0,0 +1,72 @@
+# Connecting to a node
+
+Voyager talks to your node over [Erlang distribution](https://www.erlang.org/doc/system/distributed.html), the same mechanism `iex --remsh` uses. That means two things must be true before the connect form can do anything:
+
+1. the node is **distributed**: it was started with a [name](https://www.erlang.org/doc/apps/erts/erl_cmd.html#name);
+2. Voyager knows its [**cookie**](https://www.erlang.org/doc/system/distributed.html#security).
+
+If either is missing the connection fails, usually with a vague "node down" or "connection refused". This guide covers both connection types, what each field means and what to check when it does not work.
```
srzeszut:
```suggestion
If either is missing the connection fails. This guide covers both connection types, what each field means and what to check when it does not work.
```

docs/connecting_to_a_node.md
```diff
@@ -0,0 +1,72 @@
+# Connecting to a node
+
+Voyager talks to your node over [Erlang distribution](https://www.erlang.org/doc/system/distributed.html), the same mechanism `iex --remsh` uses. That means two things must be true before the connect form can do anything:
```
srzeszut:
```suggestion
Voyager talks to your node over [Erlang distribution](https://www.erlang.org/doc/system/distributed.html). That means two things must be true before the connect form can do anything:
```

docs/connecting_to_a_node.md
```diff
@@ -0,0 +1,72 @@
...
+- **"SSH authentication failed"**: with SSH Agent, check that `ssh-add -l` lists a key the host accepts and that the agent is running. If it lists several keys of the same type, only the first is tried; see the Authentication note above. With Password, check the password. Try `ssh user@host` from a terminal first, keeping in mind that plain `ssh` tries every key and Voyager does not.
+- **Cookie mismatch**: the node logs `** Connection attempt from disallowed node ... **`. Compare the cookie with [`:erlang.get_cookie()`](https://www.erlang.org/doc/apps/erts/erlang.html#get_cookie/0) on the node.
+- **`iex --remsh` works, Voyager does not**: the short/long toggle next to the node name also decides Voyager's own name type. Set it to match how the node was started; mixed short and long names never connect.
```
srzeszut:
```suggestion
```

docs/connecting_to_a_node.md
```diff
@@ -0,0 +1,72 @@
...
+- **Advanced**: **SSH Port** (default 22) and **EPMD Port** (default 4369). Change them only when the host runs SSH or `epmd` on a non-standard port.
+
+The SSH option is greyed out when Voyager was not started with its `proxy_epmd` module. That is how the local BEAM learns to route distribution traffic through the tunnel; the packaged desktop builds always enable it.
+
+## When it does not connect
+
+- **"Node down" right away**: the node is not distributed, or the name type toggle does not match. Run `node()` in the node's shell; `:nonode@nohost` means no name was given at boot.
+- **Connection refused / timeout**: the host or the port is not reachable. For a direct connection check that `epmd` (4369) and the distribution port (random by default, pin it with [`inet_dist_listen_min` / `inet_dist_listen_max`](https://www.erlang.org/doc/apps/kernel/kernel_app.html#inet_dist_listen)) are open in the firewall. For SSH check that the SSH host itself can reach the node.
```
srzeszut:
```suggestion
- **Connection refused / timeout**: the host or the port is not reachable. For a direct connection check that `epmd` (4369) and the distribution port (random by default, you can pin it with [`inet_dist_listen_min` / `inet_dist_listen_max`](https://www.erlang.org/doc/apps/kernel/kernel_app.html#inet_dist_listen)) are open in the firewall. For SSH check that the SSH host itself can reach the node.
```

docs/connecting_to_a_node.md
```diff
@@ -0,0 +1,72 @@
...
+- **"SSH authentication failed"**: with SSH Agent, check that `ssh-add -l` lists a key the host accepts and that the agent is running. If it lists several keys of the same type, only the first is tried; see the Authentication note above. With Password, check the password. Try `ssh user@host` from a terminal first, keeping in mind that plain `ssh` tries every key and Voyager does not.
```
srzeszut:
```suggestion
- **"SSH authentication failed"**: with SSH Agent, check that `ssh-add -l` lists a key the host accepts and that the agent is running. If it lists several keys of the same type, only the first is tried; see the Authentication note above. With Password, check the password. Try `ssh user@host` from a terminal first.
```

docs/connecting_to_a_node.md
```diff
@@ -0,0 +1,72 @@
...
+- **Cookie mismatch**: the node logs `** Connection attempt from disallowed node ... **`. Compare the cookie with [`:erlang.get_cookie()`](https://www.erlang.org/doc/apps/erts/erlang.html#get_cookie/0) on the node.
+- **`iex --remsh` works, Voyager does not**: the short/long toggle next to the node name also decides Voyager's own name type. Set it to match how the node was started; mixed short and long names never connect.
+- **Name resolves differently**: a long name must resolve to the same address from Voyager and from the node. When in doubt use an IP address instead of a hostname.
+
+Recent connections are saved in a local SQLite database; secrets are encrypted before being written with a key that lives only at `~/.voyager/vault.key`.
```
srzeszut:
```suggestion
```

lib/voyager_web/live/connect_live.ex
dazakdev: Moved the icon to the right and grayed the link out when a node is connected and during the connection process.

lib/voyager_web/components/details_panel_components.ex
dazakdev: made the tooltips visible at all times, also when the data is not yet fetched

dazakdev: implemented it

piotrsac:
I think if we are making tooltips for the ETS table list headers, then the Process list should have them aswell

piotrsac:
Screenshot 2026-09-23 at 13 21 10
Screenshot 2026-09-23 at 13 20 30

The inconsistency should be fixed between processes and ETS tables list view
Im not saying it should allow the copying of the "KB" for example, i actually like the conversion to bytes

piotrsac:
@hhubert6 could you look at the suggested connecting to a node tutorial/guide? I was thinking that maybe it should show some screenshots of connect_live but on another thought it might not be necessary. What do you think?

PR #248, opened 2026-09-16
Polish list pages: process column help, stable layout, copyable protection

dazakdev:
Stacked on #245.

## What

- The processes table shows column help on hovering the header label, same as the ETS tables list in #245. Help comes from `ProcessInfoHelp`, which gets a `pid` entry.
- The summary line above the processes and ETS tables has its space reserved, so the table no longer jumps down when the first fetch lands.
- The controls label row has a fixed height on both list pages. The "?" next to Limit made the processes row taller, so its selects and table sat 7px lower than on ETS tables.
- Protection values in the ETS tables list use `value_cell`, so they get the same tooltip with a copy button as the other cells. Private tables keep their badge.

## Tests

- `details_panel_components_test.exs` asserts every process column has a help entry.


dazakdev:
Folded into #245, which now carries these commits. Closing.

PR #251, opened 2026-09-17
feat: install desktop app updates from a button

dazakdev:
Fixes VOY-113.

## What

The desktop app checks for a newer release on startup and offers it in a popup (**Update now** / **Not now**). Settings gets an **Updates** card: a dismissed update stays installable there, and **Check for updates** reports "up to date", a failed check, or turns into **Install vX** in place. Installing downloads the signed bundle, installs it and restarts the app. No silent updates: Voyager holds live connections to remote nodes, so a restart must be the user's choice.

## How

- **Rust** (`rel/app/src-tauri/src/lib.rs`): `tauri-plugin-updater`. After Elixir reports `ready` (release builds only) it calls `check()`, keeps the `Update` in state and broadcasts `available:<version>`, `none` or `check_failed` on the ElixirKit `updates` topic; it also answers a `check` request. On `install` it runs `download_and_install` + `app.restart()`, or broadcasts `failed`. A 30 s read/connect timeout turns a stalled download into `failed`.
- **Elixir**: `Voyager.Services.AppUpdater` tracks the update state (`available`, `installing`, `failed`, `checking`, `up_to_date`, `check_failed` plus a `popup?` flag set only by the startup check) and rebroadcasts it on Phoenix PubSub; `VoyagerWeb.Hooks.AppUpdateHook` assigns `@app_update` and handles the install / dismiss / check events. `Layouts.app_update_modal` is the popup, `SettingsLive.UpdateSettings` the card.
- **Config**: `plugins.updater` in `tauri.conf.json` points at `releases/latest/download/latest.json`, so pre-releases and drafts are never offered. `tauri.sh` sets `createUpdaterArtifacts` only when `TAURI_SIGNING_PRIVATE_KEY` is present, so unsigned local and `build-apps.yml` builds keep working.
- **CI** (`release.yml`): the matrix jobs get the signing secrets and no longer upload `latest.json` (each would drop the other platforms); `.github/scripts/upload_updater_json.sh` assembles it from the `.sig` assets in a final job.

## Platforms

- **macOS**: tested end to end locally (arm64): popup → install → bundle replaced → restart; dismissed update installed from Settings; stalled and corrupt downloads end in the retry state.
- **Linux AppImage**: no extra code. The plugin resolves the file from `APPIMAGE`, swaps it with a backup/rollback and `restart()` relaunches the `APPIMAGE` path. Not tested on Linux yet. Fails with a retry if the AppImage directory is read-only. Release matrix only has x86_64.

## Before marking ready

- [ ] Generate the release signing key (`cargo tauri signer generate -w <file>` with a password), store the private key and password in the team vault, and add `TAURI_SIGNING_PRIVATE_KEY` and `TAURI_SIGNING_PRIVATE_KEY_PASSWORD` as repository secrets. **Losing the key means shipped builds can never update again.**
- [ ] Replace the `pubkey` in `tauri.conf.json` with the new public key — the committed one is a local test key.
- [ ] Test the AppImage flow on Ubuntu (`dev/updater_sandbox.sh` can drive it once it lands).
- [ ] The release job (`updater_json`) has not run yet; the first tagged release verifies it.

Users on 0.1.0 have no updater, so 0.1.0 → 0.2.0 stays a manual download. When publishing, keep the pre-release flag on rc builds: only releases without it are served from `/latest`.


dazakdev:
@claude review

piotrsac:
@claude review

lib/voyager/services/app_updater.ex
```diff
@@ -0,0 +1,108 @@
...
+
+  def handle_call(:check, _from, state) do
+    ElixirKit.PubSub.broadcast(@native_topic, "check")
+
+    {:reply, :ok, put_update(state, %{version: nil, status: :checking, popup?: false})}
+  end
+
+  @impl GenServer
+  def handle_info("available:" <> version, state) do
+    popup? = not match?(%{status: :checking}, state.update)
```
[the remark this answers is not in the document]
dazakdev: szponted edge case, still wil fix

lib/voyager_web/live/settings_live/update_settings.ex
```diff
@@ -0,0 +1,83 @@
...
+                :if={!(@update && @update.status == :checking)}
+                name="icon-rotate-cw"
+                class="size-4"
+              />
+              <span
+                :if={@update && @update.status == :checking}
+                class="loading loading-spinner loading-xs"
+              ></span>
+              Check for updates
+            </button>
```
piotrsac: even if that might be a few lines of code more, i would create a new `update_button`, that would be used here two times, and in `layouts.ex` another time

piotrsac (approved):
I guess the best test of it will be in action either way 🦭

lib/voyager_web/live/settings_live/update_settings.ex
dazakdev: implemented it

PR #252, opened 2026-09-17
chore: always bind the MCP endpoint to loopback

dazakdev:
## Why

The MCP endpoint has no auth and runs tools against the connected node, so it should only ever be reachable from the same machine. `:mcp_ip` let config bind it to another interface. Nothing in the UI or README exposes that key, only the `EndpointManager` moduledoc.

It also stops making sense once #250 lands: the router will only admit loopback `Host` headers, so any non-loopback bind would answer every request with a bare 403. Two switches that can disagree, one of them silently.

## Change

- `EndpointManager` binds `127.0.0.1` unconditionally, `:mcp_ip` is no longer read, `format_host/1` is gone.
- Removed the test for the `0.0.0.0` URL mapping, the case no longer exists.
- `settings_test.exs` used `:mcp_ip` as an arbitrary key, switched to `:mcp_enabled`.

A stale `mcp_ip` row in an existing settings DB is simply never read.

Independent of #250, either can merge first.


dazakdev:
Folded into #250.

PR #253, opened 2026-09-17
Bring back default macos menu bar

piotrsac:
Im not sure why was it not used in the first place and we had to add every option other than zoom manually
Maybe there was a reason?
Its for further investigation to the reviewer


piotrsac:
The work and idea from this PR is moved into another one and will be continued

